PR #33, opened 2015-11-03
Add -r, --reverse option to reverse dependency tree

### LINKED ISSUE #25: Option for reverse mode
pipdeptree displays all packages with their dependencies.  I would like to see all packages with their dependents, to answer the question “why is X installed”.  What do you think?

tommilligan:
#### ```--reverse```

I've added ```-r``` to the command line and tried to merge it as well as I can into the existing structure. The output style is slightly different for readability, but fully compatible with the ```-f```, ```-w```, ```-a``` and ```-l``` flags.

Comments on style/string format/code welcome!

So far I've only manually tested this in 2.7 and 3.5; I haven't used tox before and have failed to get it running successfully (under Win7/MinGW), so apologies for that.

Closes #25

#### Examples:

```
$ pipdeptree -w

django-guardian==1.3.1
  - Django [installed: 1.8.5]
  - six [installed: 1.10.0]
djrill==1.4.0
  - requests [required: >=1.0.0, installed: 2.8.1]
  - Django [required: >=1.3, installed: 1.8.5]
pip-autoremove==0.9.0
virtualenv==13.1.2
WhatColorIsX==1.0.0
  - colour [installed: 0.1.2]
  - simplejson [installed: 3.8.1]
  - Pillow [installed: 3.0.0]
```
```
$ pipdeptree -w -r

virtualenv==13.1.2
Pillow==3.0.0
  - WhatColorIsX==1.0.0 [requires: Pillow]
requests==2.8.1
  - djrill==1.4.0 [requires: requests>=1.0.0]
six==1.10.0
  - django-guardian==1.3.1 [requires: six]
simplejson==3.8.1
  - WhatColorIsX==1.0.0 [requires: simplejson]
Django==1.8.5
  - djrill==1.4.0 [requires: django>=1.3]
  - django-guardian==1.3.1 [requires: Django]
colour==0.1.2
  - WhatColorIsX==1.0.0 [requires: colour]
pip-autoremove==0.9.0
```
```
$ pipdeptree -w -r -f

Django==1.8.5
    django-guardian==1.3.1
    djrill==1.4.0
pip-autoremove==0.9.0
requests==2.8.1
    djrill==1.4.0
six==1.10.0
    django-guardian==1.3.1
Pillow==3.0.0
    WhatColorIsX==1.0.0
virtualenv==13.1.2
colour==0.1.2
    WhatColorIsX==1.0.0
simplejson==3.8.1
    WhatColorIsX==1.0.0
```


tommilligan:
Issue seems to be ```render_tree()``` doesn't have some default arguments it should (e.g. ```top_pkg_str```), so I'm going to look at changing that, and also adding some ```_reverse``` tests to ```test_pipdeptree.py```

tommilligan:
```render_tree()``` now has sane default arguments (i.e. it will default to a non-frozen style output), which makes calling it during testing much more readable.

I've added some tests for ```--reverse``` as well, although obviously these may need changing if people don't like the current reversed output format.

**Comment by naiquevin** [comment]
Hi, thanks for this. I think I'll only get time to go though it over the weekend.

Testing with python 2.7 and 3.5 is fine. I'll run tox on it and do a final verification before merging the PR. Even I haven't tested under Win7/MinGW so that should be fine for now :-)

Also, when I was thinking of implementing this, my idea was to try revering the `req_map

**Comment by tommilligan** [comment]
Sorry, I should have written a bit of description. My approach was:

Parsing ```req_map``` after passing it to ```render_tree(reverse=True)``` to generate a new, ```parents``` tree structure. This basically contains the original tree structure in reverse, and can be called like ```parents[pkg.key]``` instead of ```pkg_index[pkg.key].requires()```.

After that, ```aux(reverse=True)``` is called, to indicate the ```string_generator``` used should be ```non_bottom_pkg_str``` instead of ```non_top_pkg_str```, which accepts a slightly different format (documented in the doctring), and returns a differently formatted string.

In my opinion this is better than trying to manipulate ```req_map``` beforehand, as:
* It keeps ```render_tree()``` very simple to call and use
* The same kind of reversal you suggest is pretty much what I've implemented, and it happens quietly inside ```render_tree()``` **only** if it is given the ```reverse=True``` parameter, so there is no potential slowdown for non-reversed trees
* I think the formatting of the output will need to be changed anyway for reverse trees to be intuitive

**Comment by tommilligan** [comment]
Modified the output slightly to cohere to the original style, and only show ```requires:``` if a version is specified. ```requires:``` will show both the name and version of dependency required, to avoid confusion.
```
pip-autoremove==0.9.0
colour==0.1.2
  - WhatColorIsX [installed: 1.0.0]
Werkzeug==0.10.4
  - Flask [installed: 0.10.1, requires: Werkzeug>=0.7]
    - Flask-Script [installed: 2.0.5]
Pillow==3.0.0
  - WhatColorIsX [installed: 1.0.0]
virtualenv==13.1.2
itsdangerous==0.24
  - Flask [installed: 0.10.1, requires: itsdangerous>=0.21]
    - Flask-Script [installed: 2.0.5]
simplejson==3.8.1
  - WhatColorIsX [installed: 1.0.0]
MarkupSafe==0.23
  - Mako [installed: 1.0.3, requires: MarkupSafe>=0.9.2]
  - Jinja2 [installed: 2.8]
    - Flask [installed: 0.10.1, requires: Jinja2>=2.4]
      - Flask-Script [installed: 2.0.5]
```

**Comment by naiquevin** [comment]
I can see the dict reversal you have implemented inside `render_tree` but I don't think it needs to be inside the function. If all you are doing is reversing the input, it could be done before calling the function eg. 

```
    tree = render_tree(reverse_dict(pkgs), ...)
``` 

I agree that the output formatting will probably be different for `--reverse`. But may be that can be solved by passing different formatting functions as args? In that case, it should be possible to define two high level functions eg. `render_tree_down` and `render_tree_up` that will take care of reversing the input dict and passing the correct args to `render_tree`.

IMO, multiple `if reverse then .. else ..` blocks inside `render_tree` severely affects readability of the code. I would rather have just one conditional check outside that will decide which function to delegate the job to, eg, 

```
    tree = render_tree_down(...) if not reverse else render_tree_up(...)
```

**Comment by tommilligan** [comment]
Fair enough, in that case I'll try and write something along those lines (unless you want to do it yourself)

**Comment by naiquevin** [comment]
Sure, please go ahead. I think you have already added lot of code that just needs to be moved around. Thanks.

**Comment by tommilligan** [comment]
Was trying to write as you suggested, found it quite hard without daisy-chaining ```pkg_index``` absolutely everywhere. I realised this could be easily solved with an OOP approach, and as far as I can tell it works well.

I moved everything under a ```pipdeptree.pipdeptree()``` object and found everything much easier to manage. I've removed the multiple ```if reverse: ... else:``` statements, and implemented string formatting by passing object methods through a variable, as you suggested. I also realised that you had actually already implemented a ```reverse_dict``` in the ```confusing_deps``` function :) I just hadn't realised previously. 

So, if you're interested, my last commit (e2a269e) is an OOP conversion of your original code that works on Python 2.6, 2.7, 3.3, 3.4 and 3.5. For some reason 3.2 throws an error, haven't quite worked out why yet.

Obviously this is quite a large (huge!) change in your project, so totally understand if you don't want it. I do think the OOP structure works much better though, as it allows global access to the ```pkgs```, ```pkg_index```, ```req_map```, etc data structures. It also forces testing without faking some sort of ```pkgs``` input, which I've implemented (currently running with nosetests).

Summary:
* Code is now in OOP structure, under a ```pipdeptree``` object
* Sphinx/sphinx-rtd-theme is a natural cyclic dependency, for future testing
* The output tree is now sorted alphabetically
* Backwards-compatible for command-line use

Example:
```
from pipdeptree import pipdeptree

# Main
ptd = pipdeptree(local_only=False)
tree = ptd.render_tree(list_all=True, bullets=True, reverse=False)

# Warnings
cyclic_generator = pdt.cyclic_deps()
confusing = pdt.confusing_deps()

# Helpers
nice_string = pdt.non_top_pkg_name('markupsafe', 'mako')

naiquevin:
Hi, I finally got some time to take a look at your code. To be frank, I didn't quite like the OO approach for various reasons mainly that everything is now tied together to the python environment. I can't for eg. test the functions (now methods of the class) by passing them a dict/list of packages that are not in the active python environment. The same reason why you had to modify tests that loaded pickled pkgs. Daisy chaining arguments is better IMO.

Having said that, I think both your solutions have helped me understand a key problem in my original code which is, the `render_tree` function is not abstract enough. I have a few ideas in mind for refactoring the existing code and implementing the `--reverse` flag. I am going to push my changes to a different branch so that we can compare both PRs and decide.

Thanks

tommilligan:
Sounds good. I agree that at the moment it is less flexible, but equally you could give the __init__ method an optional pickle file to use for tests, and if it is present it extracts from your test file instead of from pip; something like that? I dont think it is the OOP structure exactly that makes it inflexible.

The reason I preferred not using such a file for tests is that it feels like cheating the system a bit. If you can run and pass tests in an environment without it, what is your reason for doing so? I'm guessing reliability, in which case you could have two sets of tests, one using pickle amd the other using a real environment.

naiquevin:
Agreed, the `__init__` method could be modified to take optional args for "seeding" the object. 

Also you are right about the reason for using two kinds of tests. The very first version used pickled data stored in files. However this tool is basically a hack! It uses internal functions from the pip codebase that may change depending upon the version of pip. Pickled files gives the flexibility of simulating fake test scenarios such circular deps.

tommilligan:
```pipdeptree()``` now accepts a ```pkgs``` argument, from which all the other arguments are generated automatically. This allows for pickle-type testing. I have generated new pickle files for this testing which contain only ```pkgs``` as a list.

pkg_resources is generating a warning during testing stating that it is dropping support for Python 3.2 (which this build fails on), so I've removed it from the testing suite.

All other versions pass both pickled tests, and virtualenv-type tests. These are run in parallel and expect the same results.

naiquevin:
I have added my solution in PR https://github.com/naiquevin/pipdeptree/pull/34. It took a lot of refactoring! Basically I have defined wrapper classes for the `Distribution` and `Requirements` instances with methods such as `render_as_root` and `render_as_branch` (equivalent to top_pkg_name and non_top_pkg_name etc.) so each type of object knows how to render itself as a root node or branch node of the tree. Rest of the code is more or less the same.

Issue #20, opened 2014-07-10
Confusing Dependencies section should not display non-conflicting dependencies


thedrow:
For example:

```
* django-guardian==1.2.2 -> Django [installed: 1.6.5]
  djrill==1.0.0 -> Django [required: >=1.3, installed: 1.6.5]
  django-mptt==0.6.1 -> Django [required: >=1.4.2, installed: 1.6.5]
  django-cors-headers==0.12 -> Django [required: >=1.4, installed: 1.6.5]
```

All of them require Django above a certain version. Since none of those packages limit the Django version in use there is no conflict (Django 1.4.2 and above is required).

naiquevin:
This has been fixed in the newer version. Closing the issue. Thanks.

PR #37, opened 2015-12-04
Show warning for unsatisfied/missing dependencies

t0m:
I took a crack at #20 (using pkg_resources.parse_version to warn about unsatisfied or missing dependencies). It works pretty well on all the virtualenvs I have on my machine. 

One area that I think can be improved is where I build req_version_str via interpolation ([L307](https://github.com/naiquevin/pipdeptree/compare/master...t0m:master#diff-c9a0cc375d0f079d6251af1de64f535bR307)) so I can pass something like "pipdeptree>=0.1.2" into Requirement.parse. There must be a better way to get that string. Any suggestions?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,13 +3,14 @@ clean:
 	find . -name '*.pyc' -exec rm -f {} +
 	find . -name '*.pyo' -exec rm -f {} +
 	find . -name '*~' -exec rm -f {} +
+	find . -name '__pycache__' -exec rmdir {} +
 
 clean-env:
-	rm -rf tests/virtualenvs/{testenv,cyclicenv}
+	rm -rf tests/virtualenvs/{testenv,cyclicenv,unsatisfiedenv}
 	rm tests/virtualenvs/*.pickle
 
 test-env:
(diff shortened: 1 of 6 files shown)


naiquevin:
@t0m: thanks for this. I am a bit busy at work. Will take a look at it over the weekend.

naiquevin:
Hey sorry for the delay. I went through your code today. I think instead of having a separate warning for unsatisfied dependencies, it would make more sense to modify the "Possibly confusing dependencies" warning to show only the mismatching or conflicting deps using your code. Then it longer needs to be called "Possibly confusing". I suggest we can use the term "Conflicting" instead of "Unsatisfied" for this warning as the latter gives the impression that the dependencies of a particular project are not met, whereas pipdeptree is a standalone tool independent of any project. 

You code for finding version mismatch looks good to me. So you may just need to move that logic to `has_multi_versions` function inside `confusing_deps`.

t0m:
Hey @naiquevin, I replaced the confusing_deps function with my code and renamed it to "conflicting_deps". Let me know if that works.

PR #38, opened 2015-12-24
show full dependency requirements

chsava:
I have noticed only the base dependency requirement is being currently shown, whereas the full dependency specs are needed for this tool to be useful. Please notice requirements of pbr and SQLAlchemy below:

cristian@cristian:~/projects/pipdeptree$ pipdeptree -w -p sqlalchemy-migrate
sqlalchemy-migrate==0.10.0
  - pbr [required: >=1.3,<2.0, installed: 1.8.1]
  - six [required: >=1.7.0, installed: 1.10.0]
  - SQLAlchemy [required: >=0.7.8,!=0.9.5, installed: 1.0.11]
  - sqlparse [installed: 0.1.18]
  - decorator [installed: 4.0.6]
  - Tempita [required: >=0.4, installed: 0.5.2]

(We should also check if these dependency requirements are being currently satisfied, or if they don't conflict with those of different packages, but this will be addressed by a separate commit.)

Some minor PEP8 violations have also been fixed.

### DIFF
--- a/pipdeptree.py
+++ b/pipdeptree.py
@@ -211,12 +211,13 @@ def render_tree(tree, list_all=True, show_only=None, frozen=False):
     use_bullets = not frozen
 
     key_tree = dict((k.key, v) for k, v in tree.items())
-    get_children = lambda n: key_tree[n.key]
+
+    def get_children(n):
+        return key_tree[n.key]
 
     if show_only:
         nodes = [p for p in nodes
-                 if p.key in show_only
-                 or p.project_name in show_only]
+                 if p.key in show_only or p.project_name in show_only]
     elif not list_all:
         nodes = [p for p in nodes if p.key not in branch_keys]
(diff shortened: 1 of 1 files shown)


naiquevin:
Thanks. Before working on the conflicts part, please check an existing open pull request - https://github.com/naiquevin/pipdeptree/pull/37. I haven't had the time to have a close look at it but it seems to be addressing the same concern.

naiquevin:
PS: I don't think those are pep8 violations. See point 2 here - https://www.python.org/dev/peps/pep-0008/#other-recommendations

chsava:
You are right, the #37 commit represents a solution for displaying the conflicts. The current patch can be applied on top of it, as the full requirements specs need to be taken into account, not only their first part.

I have also updated the pep8 fix, using the latest version.

PR #39, opened 2016-01-15
add -werror to return non-zero when warnings are present

t0m:
Hi Vineet, I'd like to use pipdeptree with my buildbot builds in order to prevent any dependency issues from sneaking in over time. In order to do this, pipdeptree would have to return non-zero if there are any conflicting/cyclical dependencies to trigger a failure in the build. In order to support that I've added in a -werror flag (similar to gcc), which will return 1 when there's conflicts or cyclical deps, and 0 otherwise. I've preserved the same behavior for -w and --nowarn (to suppress the warnings), but now they cannot both be present at the same time to prevent conflicts like: `pipdeptree -werror --nowarn'. What do you think?

### DIFF
--- a/pipdeptree.py
+++ b/pipdeptree.py
@@ -350,11 +350,22 @@ def main():
                             'If in a virtualenv that has global access '
                             'donot show globally installed packages'
                         ))
-    parser.add_argument('-w', '--nowarn', action='store_true',
-                        help=(
-                            'Inhibit warnings about possibly '
-                            'confusing packages'
-                        ))
+    warn_group = parser.add_mutually_exclusive_group()
+    warn_group.add_argument('--nowarn', action='store_true',
+                            help=(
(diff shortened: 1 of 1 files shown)


naiquevin:
This is a good feature. I am thinking, instead of having two different flags we could use just one say `--warn` or `-w` with possible values - `silence` (warnings ignored) , `suppress` (warnings shown but exits with 0) and `fail` (warnings shown and exists with 1). Then `--no-warn` will go away. What do you think?

t0m:
Great! I'll try to make the changes later on today

naiquevin:
I just noticed Tavis builds are failing for reasons not related to your changes. I'll try to figure out the problem this week.

naiquevin:
Hi, I have pushed some commits on master fixing the travis-ci builds. Can you rebase your branch with my master and update the PR?

PR #43, opened 2016-04-11
ENH: dump dependency graph as graphviz dot code

johnyf:
An external call to [GraphViz `dot`](http://www.graphviz.org) is still needed, following the approach with `json`. This works even when `dot` is not installed.

An alternative would be to call the `pydot` graph method `write_pdf`, but then all layout options (including output file format) would have to be forwarded to it. Such a design would be nested, unlike the flat approach selected. It could still be added, as a `--pdf` option, or something similar.

This enhancement requires `networkx` and `pydot` (or `pydotplus`, for newer `networkx` versions).
Considering the graph layout option as optional, I did not modify the dependencies listed in `install_requires`. Both of the dependencies are pure Python, so adding them wouldn't make installation more difficult.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -201,6 +201,14 @@ tools.
 
     $ python pipdeptree.py --json
 
+The dependency graph (acyclic if there are no circular dependencies)
+can be layed out as a PDF or image using GraphViz`GraphViz
+<http://www.graphviz.org/>`_ with:
+
+.. code-block:: bash
+
+    $ pipdeptree --dot | dot -Tpdf > dependencies.pdf
+
 
 Usage
 -----
@@ -233,6 +241,7 @@ Usage
        -j, --json            Display dependency tree as json. This will yield "raw"
                              output that may be used by external tools. This option
                              overrides all other options.
+       --dot                 Print the dependency graph as GraphViz dot code.
 
 
 Known Issues
--- a/pipdeptree.py
+++ b/pipdeptree.py
@@ -7,6 +7,8 @@ import json
 
 import pip
 import pkg_resources
+# inline:
+# import networkx
 
 
 __version__ = '0.6.0'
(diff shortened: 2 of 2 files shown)


README.rst
```diff
@@ -201,6 +201,14 @@ tools.
 
     $ python pipdeptree.py --json
 
+The dependency graph (acyclic if there are no circular dependencies)
```
naiquevin: How does this change behave if there are circular dependencies?

pipdeptree.py
```diff
@@ -266,6 +268,33 @@ def jsonify_tree(tree, indent):
...
 
+def dumps_graphviz_dot(tree):
+    """Convert dependency graph to dot code for layout with graphviz.
+
+    :param dict tree: dependency graph
+    :returns: dot representation of tree
+    :rtype: str
+
+    """
+    import networkx as nx
```
naiquevin: I think you should handle `ImportError` here and fail with an error message mentioning the additional dependencies for the `--dot` flag

naiquevin:
Added a couple of inline comments. I think external call to `dot` is fine. Nested arguments would be an overkill for a small utility such as this.

johnyf:
I haven't found the time to work on these, but will eventually find it.

pipdeptree.py
johnyf: I agree, thanks to @Dunedan for addressing this in https://github.com/naiquevin/pipdeptree/pull/60/commits/0cbbee23bfc8584de240b23e6be71f16722c05c9#diff-c9a0cc375d0f079d6251af1de64f535bR356.

README.rst
johnyf: It seems to me that if the argument `tree` given to the function `dumps_graphviz_dot` is a `dict` that defines cyclic dependencies, then the resulting directed graph should contain a cycle. I can see that a warning about cyclic dependencies was added by https://github.com/Dunedan/pipdeptree/commit/ea17d2e5586e40e4f03079d9c9f450f1f25734c1#diff-c9a0cc375d0f079d6251af1de64f535bR394.

johnyf:
Thanks for merging this.

PR #52, opened 2016-09-14
Defer to pip's list of excluded default packages.

jimbocoder:
As of 6.0.0 pip will default to excluding the usual things we were
hardcoding here, so this just lets pip do its thing naturally.

This also removes 'pipdeptree' from the exclusion list, which seems
mandatory if your application actually *uses* pipdeptree and it's not
just for manual invocation (i.e.humans typing on keyboards)

### DIFF
--- a/pipdeptree.py
+++ b/pipdeptree.py
@@ -415,10 +415,7 @@ def main():
                         ))
     args = parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


naiquevin:
Thanks.

PR #60, opened 2016-12-07
Add different output options for dependency graph

Dunedan:
Add the option to output a dependency graph in any output format
supported by GraphViz.

Based on the idea and code from @johnyf to output the dependency
graph as dot code: #43

Goes a step further than that, by using `graphviz` instead of
`networkx` and `pydotplus`, which allows output as dot code as well, but
enables output in any of GraphViz supported output formats directly,
if the GraphViz command line tools are installed.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -201,6 +201,14 @@ tools.
 
     $ python pipdeptree.py --json
 
+The dependency graph (acyclic if there are no circular dependencies)
+can be layed out as a PDF or image using GraphViz`GraphViz
+<http://www.graphviz.org/>`_ with:
+
+.. code-block:: bash
+
+    $ pipdeptree --dot | dot -Tpdf > dependencies.pdf
+
 
 Usage
 -----
@@ -233,6 +241,7 @@ Usage
        -j, --json            Display dependency tree as json. This will yield "raw"
                              output that may be used by external tools. This option
                              overrides all other options.
+       --dot                 Print the dependency graph as GraphViz dot code.
 
 
 Known Issues
--- a/README.rst
+++ b/README.rst
@@ -243,7 +243,7 @@ Usage
        -j, --json            Display dependency tree as json. This will yield "raw"
                              output that may be used by external tools. This option
                              overrides all other options.
-       --output OUTPUT_FORMAT
+       --graph-output OUTPUT_FORMAT
                              Print a dependency graph in the specified output
                              format. Available are all formats supported by
                              GraphViz, e.g.: dot, jpeg, pdf, png, svg
--- a/README.rst
+++ b/README.rst
@@ -201,13 +201,15 @@ tools.
 
     $ python pipdeptree.py --json
 
-The dependency graph (acyclic if there are no circular dependencies)
-can be layed out as a PDF or image using GraphViz`GraphViz
-<http://www.graphviz.org/>`_ with:
+The dependency graph can be layed out as any of the formats supported by
+GraphViz`GraphViz<http://www.graphviz.org/>`_:
 
 .. code-block:: bash
 
-    $ pipdeptree --dot | dot -Tpdf > dependencies.pdf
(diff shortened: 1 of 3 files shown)


naiquevin:
Thanks, I will have a look at it over the weekend.

pipdeptree.py
```diff
@@ -339,6 +341,52 @@ def jsonify_tree(tree, indent):
...
+    """
+    try:
+        from graphviz import backend, Digraph
+    except ImportError:
+        print('graphviz is not available, but necessary for the output '
+              'option. Please install it.', file=sys.stderr)
+        sys.exit(1)
+
+    if output_format not in backend.FORMATS:
+        print('%s is no supported output format.' % output_format,
```
naiquevin: Can you please use str.format function for string interpolation for consistency? (I know the % syntax is used in a couple of places but I am planning to replace it with str.format)

README.rst
```diff
@@ -233,6 +243,10 @@ Usage
        -j, --json            Display dependency tree as json. This will yield "raw"
                              output that may be used by external tools. This option
                              overrides all other options.
+       --output OUTPUT_FORMAT
```
naiquevin: I think `--graph-output` would be a better name making it explicit that it will result in a graph related output

pipdeptree.py
```diff
@@ -418,6 +466,12 @@ def main():
                             '"raw" output that may be used by external tools. '
                             'This option overrides all other options.'
                         ))
+    parser.add_argument('--output', dest='output_format', required=False,
```
naiquevin: `required=False` is redundant here. Also, you may be able to use the `choices` keyword arg to validate the output formats (refer: https://docs.python.org/2.7/library/argparse.html#choices)

Dunedan: Using choices there would be awesome, but would mean to make `graphviz` a mandatory dependency as it'd be required for argument parsing in any case. I can do that, but I'm not sure if it's desired.

pipdeptree.py
```diff
@@ -339,6 +341,52 @@ def jsonify_tree(tree, indent):
...
+        print('{} is no supported output format.'.format(output_format),
+              file=sys.stderr)
+        print('Supported formats are: %s' % ', '.join(sorted(backend.FORMATS)),
+              file=sys.stderr)
+        sys.exit(1)
+
+    graph = Digraph(format=output_format)
+    for package, deps in tree.items():
+        project_name = package.project_name
+        label = '%s\n%s' % (project_name, package.version)
```
naiquevin: Sorry, I meant making the str.format change everywhere in this commit.

pipdeptree.py
```diff
@@ -418,6 +466,12 @@ def main():
                             '"raw" output that may be used by external tools. '
                             'This option overrides all other options.'
                         ))
+    parser.add_argument('--graph-output', dest='output_format',
```
naiquevin: Good point about having to add graphviz as mandatory dependency. Let's not use choices.

README.rst
```diff
@@ -201,6 +201,16 @@ tools.
 
     $ python pipdeptree.py --json
 
+The dependency graph can be layed out as any of the formats supported by
+GraphViz`GraphViz<http://www.graphviz.org/>`_:
+
+.. code-block:: bash
+
+    $ pipdeptree --output dot > dependencies.dot
```
naiquevin: There will be change here as well

naiquevin:
Merged. Thanks 👍

johnyf:
Thanks for addressing this issue. Flat is better than nested (PEP 20), so I think that using `graphviz` (equiv. `pydot`) directly, instead of via `networkx`, is a better decision.

The `graphviz` package appears to be written in pure Python, [available on PyPI](https://pypi.org/project/graphviz), and [actively maintained](https://github.com/xflr6/graphviz/commits/master), so adding it to `install_requires` shouldn't cause any installation difficulties.

PR #66, opened 2017-01-25
Fix binary GraphViz output on Python 3

jmbowman:
On Python 3, the new GraphViz PDF output failed because it was putting the `repr` of the GraphViz output bytes into the file instead of the actual bytes (so the file started with "b'%PDF" instead of "%PDF".  I fixed that and added some tests for it, making a few other fixes to running tests along the way:

* Added tests for the JSON, PDF, and SVG command line parameters and output
* Added tests for Python 3.6
* Stopped testing Python 3.2 (virtualenv dropped support for it over a year ago, making it a little hard to test with a fresh installation; Travis wasn't testing it anyway, probably for this reason)
* Added GraphViz to the Travis images and the graphviz Python package to the testing requirements
* Install tox as part of the test environment setup when running `make`
* Fixed some string formatting failures on Python 2.6
* Forced use of Python 2.7 to create the virtualenvs used to create the test data pickles; Python 3 can't create pickles of pip data that Python 2 can read even when specifying an older protocol, because some of its classes don't explicitly inherit from `object`.  These are serialized in Python 3 as new-style classes, but choke on deserialization in Python 2 because they don't have `__new__` methods there.
* Dropped wsgiref from `test_requirements.txt`.  It wasn't in use, and won't install cleanly on Python 3 even if the pickling issue above is resolved or worked around.


naiquevin:
Thanks for this. Right now I don't have the time to review the code. Will do it shortly though

PR #80, opened 2017-07-04
console-script entry point for version-specific python

ViktorHaag:
Add an entry to setup.py's entry points to create a second console-script paired to the major version of Python used to pip install. This lets us have pipdeptree console scripts installed for Python2 and Python3 for use in mixed environments.

Tested by locally installing editable via pip, and by running the tox test suite against both py27 and py36 virtual envs.

### DIFF
--- a/pipdeptree.py
+++ b/pipdeptree.py
@@ -368,7 +368,7 @@ def dump_graphviz(tree, output_format='dot'):
     graph = Digraph(format=output_format)
     for package, deps in tree.items():
         project_name = package.project_name
-        label = '{0}\n{1}'.format(project_name, package.version)
+        label = '{0} {1}'.format(project_name, package.version)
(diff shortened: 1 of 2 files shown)


ViktorHaag:
I'm unsure why `test_render_pdf` and `test_render_svg` failed on Travis; they passed locally with both testing under py27 and py36.

ViktorHaag:
It seems that PR #76 discovered similar errors in the tests, down to inserting newlines into DOT file labels, and removing the newlines rectified the situation. (Yes, DOT/Graphviz is known to be finicky about newlines.)

ViktorHaag:
Never mind. 8/

I see that a previous PR was logged to address this very issue and was rejected in favour of another solution besides alternate entrypoints. Closing and declining this PR.

Issue #100, opened 2018-05-12
Use filled node with different color for graph output


zufuliu:
Use filled node with different color for graph output, make it's clean for top level package and it's dependencies.

Changes in `dump_graphviz(tree, output_format='dot')`:
```python
    dependencies = set()
    for deps in tree.values():
        for dep in deps:
            dependencies.add(dep.project_name)
    graph = Digraph(format=output_format)
    for package, deps in tree.items():
        project_name = package.project_name
        label = '{0}\n{1}'.format(project_name, package.version)
        attrs = {'style': 'filled'}
        attrs['color'] = 'lightblue' if project_name in dependencies else 'green'
        graph.node(project_name, label=label, **attrs)
```

gaborbernat:
Seems there hasn't been any push for this for 3 years, so closing for now.

Issue #99, opened 2018-05-11
Suggest change `project_name` to lower case


zufuliu:
Suggest change `project_name` to lower case, since pip package name is case insensitive.
Packages like Pillow, Pygments and PyYAML often has different spellings in other package's dependency list. This will reduce isolated packages / dependencies produced by pipdeptree.

This change only need to add four lines in `build_dist_index`:
```python
    for p in pkgs:
        p.project_name = p.project_name.lower()
        for r in p.requires():
            r.project_name = r.project_name.lower()
    return dict((p.key, DistPackage(p)) for p in pkgs)
```

naiquevin:
Can you please open a PR with both this one and #100? Will be easier to test and merge that way.

zufuliu:
It's PR #102.

sschuberth:
This sound like a duplicate of issue https://github.com/naiquevin/pipdeptree/issues/77.

zufuliu:
@sschuberth It's a duplicate, do you have any idea to fix these failed tests in PR #102?

naiquevin:
Tests could have failed for completely different reason. The tests have become complicated and messy and I am trying to sort it out. Will test this PR and merge it. Thanks.

gaborbernat:
I think we should leave the package name normalization up to pip.

PR #102, opened 2018-05-14
PR for issue #99 and issue #100

zufuliu:
PR for issue #99 and issue #100.

### DIFF
--- a/pipdeptree.py
+++ b/pipdeptree.py
@@ -413,11 +413,14 @@ def dump_graphviz(tree, output_format='dot'):
             ', '.join(sorted(backend.FORMATS))), file=sys.stderr)
         sys.exit(1)
 
+    dependencies = set(r.key for r in flatten(tree.values()))
     graph = Digraph(format=output_format)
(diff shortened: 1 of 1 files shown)


naiquevin:
First of all, sorry for a very late reply! Finally got around to looking at the PRs on this project.

Regarding #99 : I think just making the `pkg.key` lowercase will not solve the problem because the `dump_graphviz` function is using `project_name` attribute to identify graph nodes instead of `key`.  I think a simple fix is to use `key` as the id for the graph node in the `dump_graphviz` function. It's also a "safer" solution as it won't affect rest of the code at all. 

Regarding #100 : Looks good to me.

naiquevin:
Also, don't worry about the tests. I'm working on fixing the tests in another branch. In the meanwhile, I'll manually make sure that the tests pass before merging the PR.

zufuliu:
OK, I will make rebase the code and make separate PRs later.

gaborbernat:

PR #119, opened 2019-06-11
Add 'setuptools' to list of 'install_requires'

sinoroc:
The package 'pkg_resources' is part of the 'setuptools' project.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,10 @@ with open('./README.rst') as f:
     long_desc = f.read()
 
 
-install_requires = ["pip >= 6.0.0"]
+install_requires = [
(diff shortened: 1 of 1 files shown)


naiquevin:
Hi, I just checked that `pip` also packages `pkg_resources`. So we can just import it as follows,

```
from pip._vendor import pkg_resources
```

Can you please test the above in your environment/for your use case and update the PR if it works? 

This is a good catch BTW. I think I never ran into this problem because may be all of my virtualenvs had `setuptools` installed.

sinoroc:
I see your point. For the sake of consistency, it might make sense to use the 'setuptools' vendored with 'pip' since it's here anyway.

PR #121, opened 2019-07-17
Fix missing 'setuptools' requirement

sinoroc:
The package 'pkg_resources' is part of the 'setuptools' project. The
project 'pip' (in its current version) vendors setuptools.

Since 'pipdeptree' currently uses pip's private internal API already,
we might as well use pip's vendored pkg_resources. This, of course,
might break if pip's internals change in a later version.

GitHub: naiquevin/pipdeptree#119

### DIFF
--- a/pipdeptree.py
+++ b/pipdeptree.py
@@ -19,7 +19,7 @@ try:
 except ImportError:
     from pip import get_installed_distributions, FrozenRequirement
 
-import pkg_resources
+from pip._vendor import pkg_resources
 # inline:
 # from graphviz import backend, Digraph
 


sinoroc:
It makes me a bit uneasy when I see code that relies on the internals (internal APIs, vendored libraries, etc.) of another project. It's really annoying that pip doesn't offer a stable public API. There are some other projects that can help though. setuptools / pkg_resources comes to mind obviously but there are also 'importlib_metadata', 'distlib', and some more I believe.

I had a quick look at getting rid of the dependency on 'pip' entirely. The first showstopper was the '--freeze' functionality. But this somehow seems redundant with pip's own freeze command. Maybe it's worth looking into it more seriously. Or alternatively get the pipdeptree features into pip directly.

naiquevin:
> It makes me a bit uneasy when I see code that relies on the internals (internal APIs, vendored libraries, etc.) of another project. It's really annoying that pip doesn't offer a stable public API

Yeah, this tool relies on the pip internals but that's the only way it can do what pip does (and more). One improvement on this could be `pip-shims` [1,2] which I haven't had the time to look into. For now I have just accepted that we'd need to update code and release a new version if some thing changes in pip.

The other approach is to vendor `pip` and `pkg_resources` but I don't want to go down that path. This seems to a common problem with python's build/packaging tools, specially the low level ones such as `pip` and `setuptools` - they can't depend on other libraries because if `pip` is what helps install a package resolving all it's python dependencies, how would a user install pip itself if it has any python dependencies of it's own. Probably the reason why even `pipenv` has to vendor it's dependencies [3] (but I could be wrong here)

> There are some other projects that can help though. setuptools / pkg_resources comes to mind obviously but there are also 'importlib_metadata', 'distlib', and some more I believe

`pipdeptree` aims to provide some additional functionality on top of what `pip freeze` provides. At present `pip freeze` uses `pkg_resources` to get installed distributions etc. In future that may change, and then `pipdeptree` might no longer be compatible with `pip`. So I am not in favour of using anything else directly.

> I had a quick look at getting rid of the dependency on 'pip' entirely. The first showstopper was the '--freeze' functionality. But this somehow seems redundant with pip's own freeze command

This is not correct. With the `--freeze` flag in `pipdeptree` the output is still a tree of dependencies but it's also compatible with `pip freeze` (if you ignore the indents, which pip apparently does). So you can directly write it to `requirements.txt` file and pip can still work with it at the time of installing packages. The advantage is that by reading the requirements.txt file it's easy to visually figure out which ones are top-level packages and which ones are their sub-dependencies. You can even filter only the top-level packages using `grep` before writing the output to requirements.txt (an example of this is mentioned in the README [4])

1: https://pypi.org/project/pip-shims/
2: https://github.com/naiquevin/pipdeptree/pull/106
3: https://github.com/pypa/pipenv/tree/master/pipenv/vendor
4: https://github.com/naiquevin/pipdeptree#using-pipdeptree-to-write-requirementstxt-file

sinoroc:
Hey @naiquevin,

Out of this discussion, as some kind of challenge, I had started writing an alternative implementation that doesn't rely on _pip_'s internal APIs, and eventually [deptree](https://pypi.org/project/deptree/) came out of it. It relies on _setuptools_' `pkg_resources` package instead, whose API is public and stable.

Also I would link this [_pip_ feature request](https://github.com/pypa/pip/issues/8077), since I think I had read somewhere here that you might be interested in integrating _pipdeptree_ (or at least some parts of it) in _pip_ proper.

Cheers!

PR #125, opened 2020-07-01
Avoid crashing when some dependencies aren't installed

StephenWeber:
In general, folks should have all their dependencies installed for pipdeptree to evaluate everything. However, I was using it to figure out which packages were trying to install a package that was explicitly uninstallable on my platform. This meant that pip failed mid-install, so many remaining packages had not been installed.

This change avoids a crash in the case that a dependency is not installed.

To do: perhaps it should also emit a warning? I suppose silently skipping a check is not the best for users..

### DIFF
--- a/pipdeptree.py
+++ b/pipdeptree.py
@@ -682,7 +682,7 @@ def cyclic_deps(tree):
     cyclic = []
     for p, rs in tree.items():
         for r in rs:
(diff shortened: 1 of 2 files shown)


naiquevin:
Thanks. I'll review and merge this on the weekend.

PR #132, opened 2020-09-01
Add support for extra requirements

idiot-z:
Hi
I used your code for a project where the extra requirements were important. So I added support for them.

Let me know what you think.
Best

### DIFF
--- a/pipdeptree.py
+++ b/pipdeptree.py
@@ -125,7 +125,7 @@ class DistPackage(Package):
                   tree in reverse
     """
 
-    def __init__(self, obj, req=None):
+    def __init__(self, obj, req=None, *args):
         super(DistPackage, self).__init__(obj)
         self.version_spec = None
         self.req = req
@@ -648,7 +648,7 @@ def dump_graphviz(tree, output_format='dot', is_reverse=False):
                 # particular parent package
                 req_ref = parent.req
                 edge_label = req_ref.version_spec(True) or 'any'
-                if parent.extra is None:
+                if isinstance(parent, DistPackage) or parent.extra is None:
                     graph.edge(dep.key, parent.key, label=edge_label)
                 else:
                     graph.edge(dep.key, parent.key, label=edge_label, style='dashed')
--- a/pipdeptree.py
+++ b/pipdeptree.py
@@ -151,7 +151,7 @@ class DistPackage(Package):
     def render_as_branch(self, frozen):
         assert self.req is not None
         if not frozen:
-            parent_ver_spec = self.req.version_spec
+            parent_ver_spec = self.req.version_spec()
             parent_str = self.req.project_name
             if parent_ver_spec:
                 parent_str += parent_ver_spec
@@ -180,7 +180,7 @@ class DistPackage(Package):
         """
         if req is None and self.req is None:
             return self
-        return self.__class__(self._obj, req)
+        return self.__class__(self._obj, req, req.extra if req is not None else None)
 
     def as_dict(self):
         return {'key': self.key,
@@ -201,12 +201,17 @@ class ReqPackage(Package):
     def __init__(self, obj, dist=None, extra=None):
         super(ReqPackage, self).__init__(obj)
         self.dist = dist
-        self.extra = extra
+        self._extra = extra
 
-    @property
-    def version_spec(self):
+    def version_spec(self, extra=False):
         specs = sorted(self._obj.specs, reverse=True)  # `reverse` makes '>' prior to '<'
-        return ','.join([''.join(sp) for sp in specs]) if specs else None
+        version_spec = ','.join([''.join(sp) for sp in specs]) if specs else None
+        if extra and self.extra is not None:
+            if version_spec is not None:
+                version_spec = '[{0}] {1}'.format(self.extra, version_spec)
+            else:
+                version_spec = '[{0}]'.format(self.extra)
+        return version_spec
 
     @property
     def installed_version(self):
@@ -218,12 +223,19 @@ class ReqPackage(Package):
     def is_missing(self):
         return self.installed_version == self.UNKNOWN_VERSION
 
+    @property
+    def extra(self):
+        return self._extra
+
     def is_conflicting(self):
         """If installed version conflicts with required version"""
+        # if extra requirement, no conflict
+        if self.extra is not None:
+            return False
         # unknown installed version is also considered conflicting
         if self.installed_version == self.UNKNOWN_VERSION:
             return True
-        ver_spec = (self.version_spec if self.version_spec else '')
+        ver_spec = (self.version_spec() if self.version_spec() else '')
         req_version_str = '{0}{1}'.format(self.project_name, ver_spec)
         req_obj = pkg_resources.Requirement.parse(req_version_str)
         return self.installed_version not in req_obj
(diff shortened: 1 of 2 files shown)


naiquevin:
Hi, 

First of all, sorry about the delay in looking at the PR. 

I didn't understand the need for `all`, `none` and `selection` choices for the `--extra` argument. In fact, I am not sure of the use case for having the `--extra` argument at all. IMO, pipdeptree should always consider extra requirements (at present it's a bug that it can't link extra deps). Having said that, if you have a use case, please let me know. 

I think without the `--extra` option, the implementation can be far simpler: 

1. Implement `DistPackage.extra_requires` which will return just the package keys by parsing package metadata
2. In `PackageDAG.from_pkgs` we can additionally call `p.extra_requires`, then look up the actual package object for the key from `idx` variable and call `.as_requirement()` method on it to get a ReqPackage object for it. 
3. In `ReqPackage`, we may introduce a new attribute `is_extra` to indicate that it's an extra requirement. It can be used in the presentation logic. 

```
    @classmethod
    def from_pkgs(cls, pkgs):
        pkgs = [DistPackage(p) for p in pkgs]
        idx = {p.key: p for p in pkgs}
        m = {p: [ReqPackage(r, idx.get(r.key))  for r in p.requires()] + [idx[e.key].as_requirement() for e in p.extra_requires()]
                 for p in pkgs}
        return cls(m)
```

I haven't tried this actually but the way I see it is as follows - 

Currently pipdeptree cannot identify extra requirements For eg. if `oauthlib[rsa]` is installed in the environment, then pipdeptree outputs oauthlib and cryptography as packages but cryptography is not considered a dependency of oauthlib. If we resolve this ourselves in `PackageDAG.from_pkgs`, then everything else should just work. I think the `all` and `selection` options should be automatically taken care of by the existing `--all` and `--package`/`--exclude`.

idiot-z:
Hi,
thanks for your reply.
The idea behind the `--extra` argument was that quite a few packages have a lot of extra requirements (e.g. testing, etc.) and all those extra requirements made the graphs or stdout quite unreadable. And I realized only plotting/printing those of the selection (through the `-p` option) made it clearer. Now if you do not think this is useful, I can take it out.
I'll have a look at your propositions about the implementation and come back to you.

naiquevin:
> The idea behind the --extra argument was that quite a few packages have a lot of extra requirements (e.g. testing, etc.) and all those extra requirements made the graphs or stdout quite unreadable. And I realized only plotting/printing those of the selection (through the -p option) made it clearer

yes, but won't the `-p` arg take care of that? For eg. I have installed `oauthlib[rsa]` and `ipython` in a virtualenv to test this case. `oauthlib[rsa]` installs the extra dependency `cryptography`. So,

* `pipdeptree` should show `cryptography` as a dependency of `oauthlib` (no `-p` arg)
* `pipdeptree -p oauthlib` should also show `cryptography` as a dependency of `oauthlib`. 
* But `pipdeptree -p ipython` won't  show cryptography or oauthlib 

Am I missing anything? 

> I don't understand why you differentiate: ReqPackage(r, idx.get(r.key)) and idx[e.key].as_requirement()? Should they not use the same initiation?

There's no difference actually. But with `idx[e.key].as_requirement()`, there's no need to explicitly call `pkg_resources.Requirement` which you're doing in the `extra_requires` method.

idiot-z:
Now I understand the confusion: your solution would only retain the extra dependencies that are installed on the system/virtualenv. The way I did it, all extra deps are mapped (even if not installed). I thought it'd be nice to see what is installed with which option, as prospecting if you wish. Do you think it's a bad idea?

The reason why I wrote an `extra` property and not just `is_extra` (boolean) is because the extra dependency has a name different than the parent or the dependency. E.g. wheel: [test] pytest, where test is the installation option and pytest the name of the dependency.

As for the explicit call `pkg_resources.Requirement` in the `extra_requires` method, this has to be so for the version information. E.g. for the same example as above, the way I wrote the metadata parser, it would return `('test', 'pytest (>=3.0.0)')`. Creating a `Requirement` instance from the second string in the parenthesis gives you package name and version information.

naiquevin:
> Now I understand the confusion: your solution would only retain the extra dependencies that are installed on the system/virtualenv. The way I did it, all extra deps are mapped (even if not installed). I thought it'd be nice to see what is installed with which option, as prospecting if you wish. Do you think it's a bad idea?

IMO, it's outside the scope of pipdeptree which is basically limited to only packages that are installed on the system/virtualenv. Not sure about the use case, but it could be a separate tool that provides this functionality. 

> The reason why I wrote an extra property and not just is_extra (boolean) is because the extra dependency has a name different than the parent or the dependency. E.g. wheel: [test] pytest, where test is the installation option and pytest the name of the dependency.

Got it

> As for the explicit call pkg_resources.Requirement in the extra_requires method, this has to be so for the version information. E.g. for the same example as above, the way I wrote the metadata parser, it would return ('test', 'pytest (>=3.0.0)'). Creating a Requirement instance from the second string in the parenthesis gives you package name and version information.

Makes sense. 

Also, can you please also provide the `pip freeze` output of the virtualenv where you're testing this change? I can test it against the same env.

idiot-z:
Alright, I'll leave the external packages out. Here's the pip freeze:

```
attrs==20.1.0
graphviz==0.14.1
importlib-metadata==1.7.0
iniconfig==1.0.1
more-itertools==8.5.0
packaging==20.4
pkg-resources==0.0.0
pluggy==0.13.1
py==1.9.0
pyparsing==2.4.7
pytest==6.0.1
six==1.15.0
toml==0.10.1
zipp==3.1.0
```
(pipdeptree, I installed with `-e`)

idiot-z:
I'm sorry but I can't seem to find where `test_custom_interpreter` is defined

naiquevin:
> I'm sorry but I can't seem to find where test_custom_interpreter is defined

It was introduced in a recently merged PR #128. I think you rebased your branch with latest master?

idiot-z:
Ah yes, thanks. I didn't know github performed auto rebase on the pull request.

idiot-z:
Alright, ready to merge on my side. Let me know if I can still do anything

pipdeptree.py
```diff
@@ -267,10 +306,12 @@ class PackageDAG(Mapping):
 
     @classmethod
     def from_pkgs(cls, pkgs):
-        pkgs = [DistPackage(p) for p in pkgs]
+        pkgs = [DistPackage(p) for p in pkgs if p.key != 'pkg-resources']
```
naiquevin, line 309: Why is 'pkg-resources' omitted? In case it's required, it'd be good to have a comment explaining why.  Just let me know the reason in that case, I'll add add the comment after merging the PR

idiot-z: I guess it could be left in. I forgot that I added that. The only reason I did it was that it's not even on pypi.org, I think it's being installed as part of pip. It's always alone and without dependencies and shows a version of 0.0.0 which is weird.
But I don't mind deleting that if clause if you deem it a better option.

naiquevin:
Hi, 

Thanks for your work on this long pending feature/bug. I wanted to suggest some refactoring but was finding it difficult to communicate exactly. So, I've cherry-picked your commits on another branch and added some more changes on top - https://github.com/naiquevin/pipdeptree/pull/138. Still need to get the tests passing. 

Closing this PR. It will be merged via PR #138 

Thanks

idiot-z:
Alright, perfect. Thank you too for your comments and for finishing the PR.

PR #133, opened 2020-09-21
Use Github Actions instead of Travis

gaborbernat:
@naiquevin Travis seems to not respect our pip installation commands, so moved to the more built-in Github Actions.


gaborbernat:
The action only becomes active on this repository post merge, however you can see it passing on my fork here https://github.com/gaborbernat/pipdeptree/actions/runs/264661050

naiquevin:
@gaborbernat: Wow! Thanks a lot. Now if you rebase your other PR, github action would run on it right? Can you please rebase?

naiquevin:
@gaborbernat: I missed one thing - Any idea if github actions allow us to pin the virtualenv version? For now it's working but that may change in future.

gaborbernat:
Sure I'll put in a new pr.

PR #134, opened 2020-10-08
Support --output and --exclude-regex

dldinternet:
Adding support for an output file name and using an exclude regex because listing a lot of packages by name is a pain


pipdeptree.py
```diff
@@ -376,7 +379,13 @@ def filter(self, include, exclude):
                                 continue
                 else:
                     break
-
+        assert not exclude & set([n.key for n in m])
```
naiquevin, line 382: Didn't understand why this assertion is required here.

pipdeptree.py
```diff
@@ -376,7 +379,13 @@ def filter(self, include, exclude):
...
+        if exclude_re:
+            _m = {}
+            for n, l in m.items():
```
naiquevin, line 385: IMO it's better to have this logic inside the above `for..while` loops itself instead of looping through data multiple times. To keep the code concise, you may introduce functions `is_excluded` and `is_included` within this method or as static methods of the class. For eg. `is_excluded` can take 3 args - 

1. exclude (set | None)
2. exclude_re (regex pattern | None) 
3. node key

and can return True or False. I think `exclude` (set) should get higher priority than `exclude_re` but I'm open to discussion if you think the other way around is better.

pipdeptree.py
```diff
@@ -314,7 +317,7 @@ def get_children(self, node_key):
         node = self.get_node_as_parent(node_key)
         return self._obj[node] if node else []
 
-    def filter(self, include, exclude):
+    def filter(self, include, exclude, include_re=None, exclude_re=None):
```
naiquevin, line 320: I can only see `exclude_re` used. Where is include_re being used?

pipdeptree.py
```diff
@@ -314,7 +317,7 @@ def get_children(self, node_key):
...
         """Filters nodes in a graph by given parameters
 
         If a node is included, then all it's children are also
```
naiquevin, line 323: Please update the docstring with new args

pipdeptree.py
```diff
@@ -469,7 +478,7 @@ def reverse(self):
         return PackageDAG(dict(m))
 
 
-def render_text(tree, list_all=True, frozen=False):
+def render_text(tree, list_all=True, frozen=False, output=None):
```
naiquevin, line 481: Please update docstring with new arg. Also, I suggest making it a mandatory arg. The caller can pass a file-like object explicitly.

pipdeptree.py
```diff
@@ -851,22 +873,35 @@ def main():
...
-        print_graphviz(output)
-    else:
-        render_text(tree, args.all, args.freeze)
+    if show_only is not None or exclude is not None or exclude_re is not None:
+        tree = tree.filter(show_only, exclude, include_re, exclude_re)
 
+    output = sys.stdout
+    try:
+        if args.output:
+            output = TextIOWrapper(open(args.output, 'wb'))
```
naiquevin, line 885: Can you please confirm if this works in python 2.7? Don't want to remove support for python 2.7 yet. 

I tried the following on python 2.7.14 and got an error

```
In [8]: print('abc', file=TextIOWrapper(open('/tmp/hello.txt', 'wb')))
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-8-e0a5198379f0> in <module>()
----> 1 print('abc', file=TextIOWrapper(open('/tmp/hello.txt', 'wb')))

AttributeError: 'file' object has no attribute 'readable'
```

tests/test_pipdeptree.py
```diff
@@ -66,10 +66,10 @@ def test_PackageDAG__get_node_as_parent():
 def test_PackageDAG_filter():
     # When both show_only and exclude are not specified, same tree
     # object is returned
-    assert t.filter(None, None) is t
+    assert t.filter(None, None, None, None) is t
```
naiquevin, line 69: Please add tests with non-None values

pipdeptree.py
```diff
@@ -755,6 +768,11 @@ def get_parser():
                             'Comma separated list of select packages to exclude '
                             'from the output. If set, --all will be ignored.'
                         ), metavar='PACKAGES')
+    parser.add_argument('--exclude-regex', dest='exclude_regex',
```
naiquevin, line 771: No `--include-regex' similar to `--exclude-regex'?

naiquevin (changes requested):
Thanks for the PR. I've added some inline suggestions but have some in-general comments too. 

Firstly, I think it'd be better if this change is split into 2 different PRs - (1) for `include-re/exclude-re` change and (2) for the output file arg. I think `include-re/exclude-re` is a good change and I'm looking forward to merging that part.  

But I'm unsure about the output argument change. I had considered it before but it leads to some open questions and need for additional options for the user. For eg. if there are warning about conflicts or cyclic dependencies those are written to stderr and not stdout. Should those be written to the output file as well. That's why I thought that it's better to leave out this option and have the user decide what to write to a file using bash itself 

```
pipdeptee > output.txt      # to exclude warnings in the output file

pipdeptree > output.txt 2>&1    # to include warnings in the output file 
```

Let me know what you think? 

Finally, please leave out the whitespace changes if possible as it sometimes causes unnecessary conflicts with other open PRs.

pipdeptree.py
dldinternet: It is not really required ... I added a check while making changes to "fail fast".
In fact, all assertions in the "production" code should be checks for error conditions and return values to be honest ...

pipdeptree.py
dldinternet: If you favor tidy code over performance then a helper function is fine. I will tell you that with 100's of packages the execution is already quite slow

pipdeptree.py
dldinternet: Not implemented ... TODO ...

dldinternet:
Ok I think a lot of what I you are asking makes sense. However I do not have the time to make a task out of this so perhaps I should not have made a Pull Request but rather have send you a message and said: Here is some additions I made feel free to use it.
Shall I close the PR and you can then do with the code what you want?

pipdeptree.py
dldinternet: TODO

gaborbernat:

PR #149, opened 2021-04-06
Fix GitHub Actions

pjw91:
GitHub upgraded it's `ubuntu-latest` to 20.04, which deprecated python3.4, this patch re-enables python3.4 on ubuntu18.04.

In addition, I slightly refactor the code by adding `Package.__lt__`.
As a result, package sorting can be simplified.

### DIFF
--- a/pipdeptree.py
+++ b/pipdeptree.py
@@ -119,6 +119,9 @@ class Package(object):
     def __repr__(self):
         return '<{0}("{1}")>'.format(self.__class__.__name__, self.key)
 
+    def __lt__(self, rhs):
+        return self.key < rhs.key
+
 
 class DistPackage(Package):
     """Wrapper class for pkg_resources.Distribution instances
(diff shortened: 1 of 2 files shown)


pipdeptree.py
```diff
@@ -524,6 +524,7 @@ def render_json(tree, indent):
     :rtype: str
 
     """
+    tree = tree.sort()
```
naiquevin, line 527: Any particular reason for this change?

naiquevin:
Thanks for fixing the github workflow. I'll check the other change in detail this weekend.

pipdeptree.py
pjw91: To make every run produces identical output (in Python2).

In Python2, the output changes (the order of keys is changed, thus, the hash will change) each time, which makes the output non-deterministic.

This is done in render_json_tree(), too.

naiquevin: Ok. I run integration tests[1] to see if they need to be modified and then merge the PR this weekend. Sorry about the delay. 

[1] https://github.com/naiquevin/pipdeptree/blob/master/tests/e2e-tests

naiquevin:
Thanks 👍

PR #152, opened 2021-09-01
Add test support for linux aarch64 

odidev:
Added test support for linux aarch64.in check.yml. 
@naiquevin Could you please review this PR?

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -52,3 +52,28 @@ jobs:
       run: tox -vv --notest
     - name: Run test suite
       run: tox --skip-pkg-install
+
+  build_and_test_arm64:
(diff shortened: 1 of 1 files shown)


naiquevin:
Hi @odidev, looks like you are running tests only against python 3.8 inside a docker container which is fine, but I think it's be preferable if we could use the matrix strategy similar to the existing check. I am not much familiar with github actions as well as ARM arch in general. So not sure if it's possible. Did you consider it?

naiquevin:
Just found this, but not sure if it'd be useful - https://github.com/uraimo/run-on-arch-action.

gaborbernat:
We don't plan to add a aarch64 run to CI, but if you find bugs on that platform we're happy to accept new PRS.

PR #154, opened 2021-10-03
remove_old_pip_api

stonebig:
fix https://github.com/naiquevin/pipdeptree/issues/153


naiquevin:
I've fixed the behaviour of `--user-only` and `--freeze` flags in latest commits on master. If any of you are using the `--user-only` flag, can you please test it against your env?

Will publish a new release sometime this week.

PR #156, opened 2021-11-16
improve readability

stefanm8:


naiquevin:
Didn't get the motivation behind this. That this change improves readability is very subjective IMO.

gaborbernat:
We'll say no to this for now, thanks!

PR #165, opened 2022-06-10
Add support for modified installation paths

f3flight:
This is supported by `pip freeze` argument `--path`, but currently not supported by `pipdeptree`.

### DIFF
--- a/pipdeptree.py
+++ b/pipdeptree.py
@@ -760,6 +760,10 @@ def get_parser():
                         help=(
                             'Only show installations in the user site dir'
                         ))
+    parser.add_argument('--path', action="append"
+                        help=(
+                            '[pip>=21.1] Look in a custom installation path'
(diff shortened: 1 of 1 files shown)


f3flight:
@naiquevin Hi Vineet! would you have a moment to review this simple improvement?

naiquevin:
Hi @f3flight. Thanks for the PR. It looks good on a quick glance. But I'll take some time to understand the change in more detail. I haven't been using python primarily since last few years so I am not really up-to-date with the latest developments in pip.

f3flight:
@naiquevin sure, thanks! In short, this option allows to scan an arbitrary environment/lib folders, instead of the current.

gaborbernat (changes requested):
I think the --python flag is what you're supposed to use instead of this.

PR #169, opened 2022-08-17
Issue 164, pipdeptree doesn't work with python 3.11

dqkqd:
Hi, I'm trying to fix issue https://github.com/naiquevin/pipdeptree/issues/164 which makes pipdeptree does not work with python 3.11.

Since python 3.11, pip try to use `importlib.metadata.Distribution` instead of `pkg_resources.Distribution`.
https://github.com/pypa/pip/blob/e2c9c40edeb65b9d26c13a12dbe5642dc2e00fd0/src/pip/_internal/metadata/__init__.py#L45-L49

Currently there is a workaround which is set `_PIP_USE_IMPORTLIB_METADATA=0` as mentioned in https://github.com/naiquevin/pipdeptree/issues/164#issuecomment-1145482711.
But I think it should be better to explicit import `pkg_resources` and use it.


gaborbernat (changes requested):
Please rebase on main branch.

gaborbernat (changes requested):
This PR adds 1k lines and remove 1k. In this form is just too much. We'll need to reopen a new PR that's cleanly on top of the current main.

dqkqd:
Sorry for the mess with git, I created a new one https://github.com/tox-dev/pipdeptree/pull/173

PR #171, opened 2022-09-05
Use importlib.metadata for guess_version instead of importing the module

### LINKED ISSUE #162: AssertionError with -u option
windows 10
Python 3.10.2
pip 21.3.1
pipdeptree 2.2.1

Executing `pipdeptree -u` gives the following:
```
C:\Programs\Python\lib\site-packages\_distutils_hack\__init__.py:17: UserWarning: Distutils was imported before Setuptools, but importing Setuptools also replaces the `distutils` module in `sys.modules`. This may lead to undesirable behaviors or errors. To avoid these issues, avoid using distutils directly, ensure that setuptools is installed in the traditional way (e.g. not an editable install), and/or make sure that setuptools is always imported before distutils.
  warnings.warn(
C:\Programs\Python\lib\site-packages\_distutils_hack\__init__.py:30: UserWarning: Setuptools is replacing distutils.

  warnings.warn("Setuptools is replacing distutils.")
Traceback (most recent call last):
  File "C:\Programs\Python\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Programs\Python\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\Programs\Python\Scripts\pipdeptree.exe\__main__.py", line 7, in <module>
  File "C:\Programs\Python\lib\site-packages\pipdeptree.py", line 888, in main
    conflicts = conflicting_deps(tree)
  File "C:\Programs\Python\lib\site-packages\pipdeptree.py", line 689, in conflicting_deps
    if req.is_conflicting():
  File "C:\Programs\Python\lib\site-packages\pipdeptree.py", line 231, in is_conflicting
    if self.installed_version == self.UNKNOWN_VERSION:
  File "C:\Programs\Python\lib\site-packages\pipdeptree.py", line 221, in installed_version
    return guess_version(self.key, self.UNKNOWN_VERSION)
  File "C:\Programs\Python\lib\site-packages\pipdeptree.py", line 64, in guess_version
    m = import_module(pkg_key)
  File "C:\Programs\Python\lib\importlib\__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "C:\Programs\Python\lib\site-packages\setuptools\__init__.py", line 8, in <module>
    import _distutils_hack.override  # noqa: F401
  File "C:\Programs\Python\lib\site-packages\_distutils_hack\override.py", line 1, in <module>
    __import__('_distutils_hack').do_override()
  File "C:\Programs\Python\lib\site-packages\_distutils_hack\__init__.py", line 71, in do_override
    ensure_local_distutils()
  File "C:\Programs\Python\lib\site-packages\_distutils_hack\__init__.py", line 59, in ensure_local_distutils
    assert '_distutils' in core.__file__, core.__file__
AssertionError: C:\Programs\Python\lib\distutils\core.py
```
Thanks

saaketp:
Fixes #162 

The current approach tries to import the package and look at `package.__version__`. This is not exactly reliable and also causes Assertion Error in the case where the packages is `setuptools` (because pipdeptree imports `pip` and `setuptools` can't be imported after `pip` because of weird hacks in setuptools). The better and simpler approach is to use `importlib.metadata.version`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ optional-dependencies.test = [
   "pytest>=7.1.3",
   "pytest-cov>=3",
   "pytest-mock>=3.8.2",
-  "virtualenv<21,>=20.16.4",
+  "virtualenv<21,>=20.16.5",
 ]
 optional-dependencies.graphviz = [
   "graphviz>=0.20.1",
--- a/src/pipdeptree/__init__.py
+++ b/src/pipdeptree/__init__.py
@@ -42,6 +42,18 @@ def guess_version(pkg_key, default="?"):
     :rtype: string
     """
     try:
+        if sys.version_info >= (3, 8):  # pragma: >=3.8 cover
+            import importlib.metadata as importlib_metadata
+        else:  # pragma: <3.8 cover
+            import importlib_metadata
+        return importlib_metadata.version(pkg_key)
+    except ImportError:
+        pass
+    # Avoid AssertionError with setuptools
+    # see https://github.com/tox-dev/pipdeptree/issues/162
+    if pkg_key in {"setuptools"}:
+        return default
+    try:
         m = import_module(pkg_key)
     except ImportError:
         return default
--- a/src/pipdeptree/__init__.py
+++ b/src/pipdeptree/__init__.py
@@ -49,8 +49,7 @@ def guess_version(pkg_key, default="?"):
         return importlib_metadata.version(pkg_key)
     except ImportError:
         pass
-    # Avoid AssertionError with setuptools
-    # see https://github.com/tox-dev/pipdeptree/issues/162
+    # Avoid AssertionError with setuptools, see https://github.com/tox-dev/pipdeptree/issues/162
     if pkg_key in {"setuptools"}:
         return default
     try:
--- a/tests/test_pipdeptree.py
+++ b/tests/test_pipdeptree.py
@@ -11,7 +11,6 @@ except ImportError:
     from unittest import mock
 
 import pytest
-import virtualenv
 
 import pipdeptree as p
 
@@ -466,6 +465,8 @@ def test_parser_svg():
 
 @pytest.mark.parametrize("args_joined", [True, False])
 def test_custom_interpreter(tmp_path, monkeypatch, capfd, args_joined):
+    import virtualenv
+
     result = virtualenv.cli_run([str(tmp_path), "--activators", ""])
     cmd = [sys.executable]
     cmd += [f"--python={result.creator.exe}"] if args_joined else ["--python", str(result.creator.exe)]
@@ -495,7 +496,6 @@ def test_guess_version_setuptools_not_imported(monkeypatch):
     builtin_import = builtins.__import__
 
     def _import(name, *args, **kwargs):
-        assert name not in {"setuptools"}
         if name in {"importlib.metadata", "importlib_metadata"}:
             raise ImportError
         return builtin_import(name, *args, **kwargs)
--- a/tests/test_pipdeptree.py
+++ b/tests/test_pipdeptree.py
@@ -1,4 +1,3 @@
-import builtins
 import platform
 import sys
 from contextlib import contextmanager
@@ -491,14 +490,10 @@ def test_custom_interpreter(tmp_path, monkeypatch, capfd, args_joined):
     assert err == "graphviz functionality is not supported when querying" " non-host python\n"
 
 
-def test_guess_version_setuptools_not_imported(monkeypatch):
-    builtin_import = builtins.__import__
-
-    def _import(name, *args, **kwargs):
-        if name in {"importlib.metadata", "importlib_metadata"}:
-            raise ImportError
-        return builtin_import(name, *args, **kwargs)
-
-    with monkeypatch.context() as m:
-        m.setattr(builtins, "__import__", _import)
-        assert p.guess_version("setuptools") == p.ReqPackage.UNKNOWN_VERSION
+def test_guess_version_setuptools_not_imported(mocker):
+    mock_importlib_metadata = mocker.Mock(spec_set=["version"])
+    mock_importlib_metadata.version = mocker.Mock(side_effect=ImportError)
+    mocker.patch.dict("sys.modules", {"importlib.metadata": mock_importlib_metadata})
+    mock_import_module = mocker.patch.object(p, "import_module", autospec=True)
+    assert p.guess_version("setuptools") == p.ReqPackage.UNKNOWN_VERSION
+    mock_import_module.assert_not_called()
--- a/tests/test_pipdeptree.py
+++ b/tests/test_pipdeptree.py
@@ -491,9 +491,11 @@ def test_custom_interpreter(tmp_path, monkeypatch, capfd, args_joined):
 
 
 def test_guess_version_setuptools_not_imported(mocker):
+    mock_importlib = mocker.Mock(spec_set=["metadata"])
     mock_importlib_metadata = mocker.Mock(spec_set=["version"])
+    mock_importlib.metadata = mock_importlib_metadata
     mock_importlib_metadata.version = mocker.Mock(side_effect=ImportError)
-    mocker.patch.dict("sys.modules", {"importlib.metadata": mock_importlib_metadata})
+    mocker.patch.dict("sys.modules", {"importlib": mock_importlib, "importlib.metadata": mock_importlib_metadata})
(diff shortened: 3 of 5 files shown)


src/pipdeptree/__init__.py
```diff
@@ -61,15 +60,13 @@ def guess_version(pkg_key, default="?"):
 
     """
     try:
-        m = import_module(pkg_key)
+        if sys.version_info >= (3, 8):
```
gaborbernat, line 63: Add the covdefaults marker to allow this if/else to be covered correcvtly.

pyproject.toml
```diff
@@ -16,6 +16,9 @@ urls.Homepage = "https://github.com/tox-dev/pipdeptree"
 urls.Source = "https://github.com/tox-dev/pipdeptree"
 urls.Tracker = "https://github.com/tox-dev/pipdeptree/issues"
 requires-python = ">=3.7"
+dependencies = [
+  "importlib_metadata; python_version < '3.8'"
```
gaborbernat, line 20: We cannot take on any dependencies, as that would mean the `--python` flag would fail if the target python doesn't have the given dependency installed.

gaborbernat (changes requested):

src/pipdeptree/__init__.py
saaketp: Added

pyproject.toml
saaketp: It will not really fail because we have `except ImportError: return default`, it will just show the version as `?`.

And the target venv probably has it installed anyway as a transitive dependency, too many packages depend on it (like pytest).

We can still remove the dependency, but then what to do for 3.7?
I can think of few ways to handle it:
1) Current behavior (in this branch): Just try importing `importlib_metadata` but without adding it as a dependency and fallback to default `?` version if it can't be imported.
2) Fall back to the `package.__version__` heuristic that is implemented in the main branch, with some special casing for `setuptools` to avoid the `AssertionError`.
3) Report `?` unconditionally
4) Drop support for python 3.7

gaborbernat: We need to do 2 realistically.

saaketp: Okay so the messy one that I was trying to avoid :)

I had to force-push to handle the merge conflicts anyway, so I started from scratch, and I think I have a cleaner way now.

The `package.__version__` way is worse than using `importlib_metadata` so I am still trying `importlib_metadata` first, but if that fails then the old logic runs (except in case of setuptools, where we just return `?`).

src/pipdeptree/__init__.py
```diff
@@ -41,6 +41,18 @@ def guess_version(pkg_key, default="?"):
...
+            import importlib.metadata as importlib_metadata
+        else:  # pragma: <3.8 cover
+            import importlib_metadata
+        return importlib_metadata.version(pkg_key)
+    except ImportError:
+        pass
+    # Avoid AssertionError with setuptools
+    # see https://github.com/tox-dev/pipdeptree/issues/162
+    if pkg_key in {"setuptools"}:
+        return default
```
gaborbernat, line 55: Seems this line is not covered in the test suite, also move the comment into one line

gaborbernat (changes requested):

src/pipdeptree/__init__.py
saaketp: So this case triggers only if we have `--local-only` or `--user-only` command line flag, and `setuptools` is not installed in the current venv or the user site packages (possibly available through the system site packages). But there is some package in the venv dependent on `setuptools`, like `pytest` or `ipython`. And I'm not sure how to create such an environment when the tests run.

saaketp: Or well maybe we just need some package dependent on setuptools in the venv, but don't have setuptools installed.

The `--user-only` and `--local-only` thing is only needed to trigger the `AssertionError`.

saaketp: And I just realized I don't need to make a real venv to test it anyway, let me give it a try.

saaketp: Okay I have added a simple test.

I wanted to replicate the AssertionError too, but it was not simple because the test imports `virtualenv`, which imports `distutils` and that applies the `setuptools` hack before `pipdeptree` gets imported. The test that relies on `virtualenv` would start failing if we try to import `pipdeptree` before `virtualenv`.

saaketp: Actually I was being stupid, `virtualenv` imports `disutils` not `setuptools` so that test doesn't really fail if we move the `virtualenv` import below.

So now I have updated the test, so that it would give AssertionError without the changes from this PR.

tests/test_pipdeptree.py
```diff
@@ -488,3 +490,16 @@ def test_custom_interpreter(tmp_path, monkeypatch, capfd, args_joined):
...
+
+def test_guess_version_setuptools_not_imported(monkeypatch):
+    builtin_import = builtins.__import__
+
+    def _import(name, *args, **kwargs):
+        if name in {"importlib.metadata", "importlib_metadata"}:
+            raise ImportError
+        return builtin_import(name, *args, **kwargs)
+
+    with monkeypatch.context() as m:
```
gaborbernat, line 503: use pytes-tmock instead of monkeypatch - https://github.com/tox-dev/pipdeptree/blob/main/pyproject.toml#L25

tests/test_pipdeptree.py
```diff
@@ -488,3 +490,16 @@ def test_custom_interpreter(tmp_path, monkeypatch, capfd, args_joined):
     assert context.value.code == 1
     assert not out
     assert err == "graphviz functionality is not supported when querying" " non-host python\n"
+
+
+def test_guess_version_setuptools_not_imported(monkeypatch):
+    builtin_import = builtins.__import__
+
+    def _import(name, *args, **kwargs):
```
gaborbernat, line 498: This should match exactly the interface of that method and not use args, kwargs

tests/test_pipdeptree.py
```diff
@@ -465,6 +465,8 @@ def test_parser_svg():
 
 @pytest.mark.parametrize("args_joined", [True, False])
 def test_custom_interpreter(tmp_path, monkeypatch, capfd, args_joined):
+    import virtualenv
```
gaborbernat, line 468: This should be at the top of the file

gaborbernat (changes requested):

tests/test_pipdeptree.py
saaketp: I moved it down to get the AssertionError from setuptools that was reported in #162.
I moved it up again but now the test passes even without the changes in the module from this PR.

tests/test_pipdeptree.py
saaketp: Got rid of it with the pytest-mock change

tests/test_pipdeptree.py
saaketp: Done

tests/test_pipdeptree.py
gaborbernat: Hmm, then we'll need to change it around, but this test sounds very brittle :thinking: perhaps rework it into a subprocess call to not make it depend on setuptools being imported?

saaketp: Yeah it is a weird one.
I gave it a try with subprocess now, how does that look?

tests/test_pipdeptree.py
```diff
@@ -488,3 +490,25 @@ def test_custom_interpreter(tmp_path, monkeypatch, capfd, args_joined):
...
+    else:
+        import importlib_metadata
+
+    def raise_import_error(name):
+        raise ImportError(name)
+
+    importlib_metadata.version = raise_import_error
+
+    print(pipdeptree.guess_version("setuptools"), end="")
+    """
```
gaborbernat, line 512: You loose code coverage and debug-ability this way. Make it a side file instead, and call the python file via sys.executable instead of -c

gaborbernat (changes requested):

gaborbernat (approved):

PR #178, opened 2022-10-01
Update README for tested Python versions

tomkins:
Mostly just removing older versions and adding 3.10.

I've removed the comment about 2.6 - support for that ended nearly 9 years ago!

Feel free to edit/tweak as needed.


nohamelin:
Hi

I understand removing any mention of 2.x, but it should be replaced instead with a mention about the last release tested with 3.6/3.5

Issue #188, opened 2023-01-09
Graph stability: random order in Graphiz content


kdeldycke:
In the same environment and the same set of dependency, the output generated by `pipdeptree` is not the same.

To test this, run in succession the same `pipdeptree` call:
```shell-session
$ pipdeptree --packages click-extra --graph-output dot > graph1.dot
$ pipdeptree --packages click-extra --graph-output dot > graph2.dot
$ pipdeptree --packages click-extra --graph-output dot > graph3.dot
```

You can check none of the produced files are identical:
```shell-session
❯ shasum graph*.dot
302894f16f812b6e360e2360198f90d1d187c6a9  graph1.dot
8e8de92b7eed4d95d7385a53e3ab570ce8101bf5  graph2.dot
c283579a78573ae431961dc19aec09d752c8cabf  graph3.dot
```

That's because the edges produced in the output are not sorted and randomly ordered at each `pipdeptree` invokation:
```diff
--- ./graph1.dot	2023-01-09 15:46:19
+++ ./graph2.dot	2023-01-09 15:46:23
@@ -1,85 +1,85 @@
 digraph {
 	"click-extra" [label="click-extra\n3.8.0"]
-	"click-extra" -> xmltodict [label=">=0.12,<0.14"]
-	"click-extra" -> commentjson [label=">=0.9.0,<0.10.0"]
-	"click-extra" -> regex [label=">=2022.3.15,<2023.0.0"]
+	"click-extra" -> mergedeep [label=">=1.3.4,<2.0.0"]
 	"click-extra" -> cloup [label=">=2.0.0.post1,<3.0.0"]
-	"click-extra" -> wcmatch [label=">=8.4,<9.0"]
-	"click-extra" -> sphinx [label=">=5.3.0,<6.0.0"]
-	"click-extra" -> "pygments-ansi-color" [label=">=0.0.6,<0.1.1"]
-	"click-extra" -> pygments [label=">=2.14.0,<3.0.0"]
 	"click-extra" -> click [label=">=8.1.1,<9.0.0"]
-	"click-extra" -> "pallets-sphinx-themes" [label=">=2.0.2,<3.0.0"]
+	"click-extra" -> pygments [label=">=2.14.0,<3.0.0"]
+	"click-extra" -> "pygments-ansi-color" [label=">=0.0.6,<0.1.1"]
 	"click-extra" -> requests [label=">=2.27.1,<3.0.0"]
[...]

PR #189, opened 2023-01-09
Make the output of the dot format deterministic and stable

kdeldycke:
This is an adhoc re-interpretation of the [`graphviz.dot.Dot.__iter__()` method](https://github.com/xflr6/graphviz/blob/master/graphviz/dot.py#L158-L182) that is used behind the scene by [the original `graph.source` call](https://github.com/tox-dev/pipdeptree/blob/2eb489449169b59711a4fc979f4fc3483e8e81b0/src/pipdeptree/__init__.py#L613-L615) to dump the graphviz dot file content in `pipdeptree` (i.e. when the `--graph-output dot` parameter in called on the CLI).

But in this PR, we forces the sorting of the generated nodes and edge statements of the graph. That way we address issue #188, in which each invokation of the `pipdeptree` generates a different dot file.

The order of the `dot`-statements does not matter because of the nature of dependencies and the way `pipdeptree` process them. That's why a call to `sorted()` works without any special care. Graphviz will anyway re-interpret internally the relationships on rendering.

### DIFF
--- a/src/pipdeptree/__init__.py
+++ b/src/pipdeptree/__init__.py
@@ -612,7 +612,11 @@ def dump_graphviz(tree, output_format="dot", is_reverse=False):
 
     # Allow output of dot format, even if GraphViz isn't installed.
     if output_format == "dot":
-        return graph.source
+        # Emulates graphviz.dot.Dot.__iter__() to force the sorting of graph.body.
+        # Fixes https://github.com/tox-dev/pipdeptree/issues/188
+        # That way we can guarantee the output of the dot format is deterministic
+        # and stable.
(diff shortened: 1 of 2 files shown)


gaborbernat (changes requested):
Fix CI, rebase on main and add test.

kdeldycke:
Just rebased, squashed, fixed and added unit-tests to ensure stability of Graphviz output.

All tests are passing, this PR is ready to be merged upstream.

gaborbernat (approved):

kdeldycke:
Thanks @gaborbernat for the merge!

Issue #207, opened 2023-03-17
AttributeError messages generating mermaid output for reverse dependencies


ajkerrigan:
I was using a command like this (on pipdeptree version `2.5.2`):

```console
❯ pipdeptree --python (py -c 'import sys; print(sys.executable)') -p packaging -r --mermaid
```

And getting these errors:

```console
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/tmp/tmpyyjztxs5/pipdeptree/__main__.py", line 6, in <module>
    sys.exit(main())
             ^^^^^^
  File "/tmp/tmpyyjztxs5/pipdeptree/__init__.py", line 980, in main
    print(render_mermaid(tree))
          ^^^^^^^^^^^^^^^^^^^^
  File "/tmp/tmpyyjztxs5/pipdeptree/__init__.py", line 609, in render_mermaid
    pkg_label = f"{pkg.project_name}\\n{pkg.version}"
                                        ^^^^^^^^^^^
  File "/tmp/tmpyyjztxs5/pipdeptree/__init__.py", line 121, in __getattr__
    return getattr(self._obj, key)
           ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Requirement' object has no attribute 'version'
```

```console
Traceback (most recent call last):                                                                                                                                
  File "<frozen runpy>", line 198, in _run_module_as_main                                                                                                         
  File "<frozen runpy>", line 88, in _run_code                                                                                                                    
  File "/tmp/tmplc94do0z/pipdeptree/__main__.py", line 6, in <module>                                                                                             
    sys.exit(main())                                                                                                                                              
             ^^^^^^                                                                                                                                               
  File "/tmp/tmplc94do0z/pipdeptree/__init__.py", line 980, in main                                                                                               
    print(render_mermaid(tree))                                                                                                                                   
          ^^^^^^^^^^^^^^^^^^^^                                                                                                                                    
  File "/tmp/tmplc94do0z/pipdeptree/__init__.py", line 615, in render_mermaid                                                                                     
    if dep.is_missing:                                                                                                                                            
       ^^^^^^^^^^^^^^                                                                                                                                             
  File "/tmp/tmplc94do0z/pipdeptree/__init__.py", line 121, in __getattr__                                                                                        
    return getattr(self._obj, key)                                                                                                                                
[...]

PR #208, opened 2023-03-17
Fix attribute errors for reverse dependency mermaid output

ajkerrigan:
I'm not sure this is the _right_ fix, but it's something that seemed to at least dodge attribute errors locally and produce the output I was expecting. So maybe if nothing else it's an input that might lead to the right fix :).

(Details about the problem this is trying to solve in #207)

### DIFF
--- a/src/pipdeptree/__init__.py
+++ b/src/pipdeptree/__init__.py
@@ -606,10 +606,7 @@ def render_mermaid(tree) -> str:
     edges: set[str] = set()
 
     for pkg, deps in tree.items():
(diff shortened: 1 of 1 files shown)


gaborbernat (changes requested):
Test please.

ajkerrigan:
That's fair. Eyeballing closer this clearly isn't the right answer anyway, not picking up the versions. I'm going to have to ramp up on context to get something together that's worth a real review rather than being a conversation starter.

Will close this in the meantime and reopen if/when I have something useful. Thanks.

Issue #213, opened 2023-03-22
Feature proposal: Use fnmatch for include and exclude


jensens:
## Problem

Here in our project we have many packages in the same namespace to exclude. The idea is to get a dependency tree without noise from the framework, but focus on the own packages. This results in a very long exclude list, like 

```
zope.configuration,zope.deferredimport,zope.publisher,zope.testing,zope.annotation,zope.sequencesort,zope.structuredtext,zope.componentvocabulary,zope.dottedname,zope.lifecycleevent,zope.pagetemplate,zope.traversing,zope.viewlet,zope.intid,zope.keyreference,zope.site,zope.globalrequest,zope.browserresource,zope.ramcache,zope.browser,zope.container,zope.browserpage,zope.contentprovider,zope.filerepresentation,zope.size,zope.cachedescriptors,zope.copy,zope.datetime,zope.browsermenu,zope.app.locales,zope.sendmail,zope.contenttype,zope.exceptions,zope.processlifetime,zope.ptresource,zope.tales,zope.testbrowser
``` 

## Proposal

I propose to use Pythons [`fnmatch`](https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatch) to check if a package is in the given set of excludes. I propose to apply the same on includes to get a consistent behavior. In this case the above list can be reduced to

```
zope.*
```

## Proposed implementation

Affected are the current code blocks here:
1. https://github.com/tox-dev/pipdeptree/blob/b15e50abbba7415313b36076dd83588c34220a8d/src/pipdeptree/__init__.py#L353-L356
2. https://github.com/tox-dev/pipdeptree/blob/b15e50abbba7415313b36076dd83588c34220a8d/src/pipdeptree/__init__.py#L360

Define a general multi-pattern-checker:

```python
import fnmatch 

def multi_fnmatch(key, patterns):
    for pattern in pattern:
        if fnmatch.fnmatch(key, pattern)
            return True
    return False
```

Replace current first code block with

```python
            if multi_fnmatch(node.key, exclude):
                continue
            if include is None or multi_fnmatch(node.key, include):
                stack.append(node)
```
and replace current second code block with
```
                    cldn = [c for c in self._obj[n] if not multi_fnmatch(c.key, exclude)]
```

At the moment there are no tests for the code covered by the change. 
I propose to add at least a simple test for the new `multi_fnmatch`. 

## Plan for implementation

I can implement this feature.

- wait for feedback. 
[...]

PR #215, opened 2023-03-24
Implement: Use fnmatch for include and exclude #213

jensens:
see proposal in #213 for details

### DIFF
--- a/src/pipdeptree/__init__.py
+++ b/src/pipdeptree/__init__.py
@@ -1,4 +1,5 @@
 import argparse
+import fnmatch
 import inspect
 import json
 import os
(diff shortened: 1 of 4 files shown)


src/pipdeptree/__init__.py
```diff
@@ -89,6 +90,14 @@ def frozen_req_from_dist(dist):
         return FrozenRequirement.from_dist(dist, [])
 
 
+def multi_fnmatch(key, patterns):
```
gaborbernat, line 93: Use `any` instead with a generator expression and inline it

gaborbernat (changes requested):

gaborbernat (approved):

PR #254, opened 2023-06-29
Added new argument that lets you manually override the unicode detection for cleaner logging purposes.

dparkinson101:
Included argument that lets you manually override the encoding type and added basic test to make sure it works.


src/pipdeptree/__init__.py
```diff
@@ -997,6 +997,15 @@ def get_parser():
...
     )
+    parser.add_argument(
+        "--without-unicode",
+        action="store_true",
+        default=False,
+        help=(
+            "Display dependency tree as text without using unicode characters."
+            "This option overrides default unicode detection."
+        ),
+    )
```
gaborbernat, line 1008: I'd rather make this a flag of encoding; that the user can set to ascii if they don't want the utf characters 👍

gaborbernat (changes requested):

gaborbernat (changes requested):
Please also include a test.

src/pipdeptree/__init__.py
```diff
@@ -997,6 +1000,15 @@ def get_parser():
             " ignore this argument."
         ),
     )
+    parser.add_argument(
+        "--encoding",
+        dest="encoding_type",
+        default=None,
```
gaborbernat, line 1006: Shouldn't thje default here be sys.stdout.encoding.lower(), because that's what we're defaulting to and render_text should not take an optional arg, should be mandatory.

src/pipdeptree/__init__.py
```diff
@@ -997,6 +1000,15 @@ def get_parser():
...
+        help=(
+            "Display dependency tree as text using specified encoding (defaults to ascii)"
```
gaborbernat, line 1008: Please align to 120 characters.

gaborbernat (changes requested):

gaborbernat:
Superseeded by https://github.com/tox-dev/pipdeptree/pull/255

PR #267, opened 2023-07-15
Simplify loops

eumiro:
(the previous PR got closed because I accidentally deleted the branch)


eumiro:
The `mermaid_id` method always returns a `str`, so there's no need to return something else at the end, but the checks fail. What would you like to return there? A non-existent `str` or flag the method to be skipped by the test?

gaborbernat (approved):

PR #279, opened 2023-08-14
Exit with error when `-p` given patterns that fail to match

### LINKED ISSUE #127: Success code returned on missing packages
I was totally confused by output of pipdeptree when I tried to call it with `pipdeptree -p ansiblelint`, as it listed not deps and returned as success.

It was my mistake to mention module name instead of correct packaga name `ansible-lint` but clearly returning 0 is a bug, as a missing package should clearly return an error.

kemzeb:
Should resolve #127.

Note that the `include` parameter in `filter_nodes()` was turned into a list so that when printing an error they will appear in order (at the cost of allowing duplicate patterns to be added). I don't see a performance cost by doing this since previously we seem to just iterate `include` rather than do any lookups.

When the -w fail option and the --packages option is given, patterns that fail to match with any package names will printed to stderr and return with exit code 1.

-w suppress and -w silence work as attended.

### DIFF
--- a/src/pipdeptree/_models/dag.py
+++ b/src/pipdeptree/_models/dag.py
@@ -91,14 +91,15 @@ class PackageDAG(Mapping[DistPackage, List[ReqPackage]]):
         node = self.get_node_as_parent(node_key)
         return self._obj[node] if node else []
 
-    def filter_nodes(self, include: set[str] | None, exclude: set[str] | None) -> PackageDAG:  # noqa: C901
+    def filter_nodes(self, include: list[str] | None, exclude: set[str] | None) -> PackageDAG:  # noqa: C901
         """
         Filters nodes in a graph by given parameters.
 
         If a node is included, then all it's children are also included.
 
-        :param include: set of node keys to include (or None)
+        :param include: list of node keys to include (or None)
         :param exclude: set of node keys to exclude (or None)
+        :raises ValueError: If include has node keys that do not exist in the graph
         :returns: filtered version of the graph
         """
         # If neither of the filters are specified, short circuit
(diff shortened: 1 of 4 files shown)


kemzeb:
Recent change removes an unnecessary branch. Not sure why the tests are failing though. It seems that the required test coverage is 88%, but it fails because we are under 100%?

src/pipdeptree/_models/dag.py
```diff
@@ -110,25 +111,40 @@ def filter_nodes(self, include: set[str] | None, exclude: set[str] | None) -> Pa
...
-            include = {s.lower() for s in include}
+            include_with_casing_preserved = include
+            include = [s.lower() for s in include]
         exclude = {s.lower() for s in exclude} if exclude else set()
 
         # Check for mutual exclusion of show_only and exclude sets
         # after normalizing the values to lowercase
         if include and exclude:
-            assert not (include & exclude)
+            assert not (set(include) & exclude)
```
kemzeb, line 123: A tad off-topic, but do we need these lines? `--package` and `--exclude` aren't able to be used together anyways so there should never be a case where `include` and `exclude` will both be non-empty iterables

kemzeb: Oh, after reading the CI logs I believe the CI failures (at least for testing) have to do with these lines since we have no tests that cover this code path (specifically we don't cover line 123). Not sure how there hasn't been complaints in previous commits to main though.

gaborbernat: I think is that all new changes must have coverage... even if they did not have before.

kemzeb: Should I remove these lines (and maybe raise a `ValueError` as well as document an invariant explaining that `include` and `exclude` cannot both have values) or would you rather I add a test case to cover them?

gaborbernat: Add a test to cover them 👍

kemzeb:
This comment ignores the ruff rule PLR0912 in `filter_nodes()` so that I can have branches > 12. Let me know if this is a problem

PR #289, opened 2023-09-27
precommit: enable pyupgrade with `py3.8`+

Borda:
adding automatic syntax upgrade to python 3.8 or higher :rabbit: 
with running in on the repo seems all is fine, so this will just hold the bar

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,9 +1,24 @@
+default_language_version:
+  python: python3
+
+ci:
+  autofix_prs: true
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -1,9 +1,24 @@
+default_language_version:
```
gaborbernat, line 1: Covered by ruff already. https://github.com/tox-dev/pipdeptree/blob/main/pyproject.toml#L71

gaborbernat (changes requested):

PR #290, opened 2023-09-27
precommit: enable docformatter

Borda:
This tool helps to update docs and re-enable Ruff `D` with some fixes to be compatible

also resolving issue flagged by Ruff and ignored


pyproject.toml
```diff
@@ -107,3 +106,10 @@ overrides = [
     "virtualenv.*",
   ], ignore_missing_imports = true },
 ]
+
+[tool.docformatter]
+recursive = true
+# this need to be shorter as some docstings are r"""...
+wrap-summaries = 119
+wrap-descriptions = 120
+blank = true
```
gaborbernat, line 115: Why, isn't ruff already doing this?

gaborbernat (changes requested):

pyproject.toml
Borda: First, it has been disabled, see:
https://github.com/tox-dev/pipdeptree/blob/71c92458e1427a8b4fd0f88746b1e50059c950f6/pyproject.toml#L76
and second from personal experience `docformatter` does better job with line lengths

.pre-commit-config.yaml
```diff
@@ -27,7 +33,7 @@ repos:
     rev: "v0.0.291"
     hooks:
       - id: ruff
-        args: [--fix, --exit-non-zero-on-fix]
+        args: ["--fix"]
```
Borda, line 36: this is already handled by pre-commit as any change because of `--fix` yield in error

pyproject.toml
```diff
@@ -72,10 +72,8 @@ target-version = "py38"
 isort = {known-first-party = ["pipdeptree"], required-imports = ["from __future__ import annotations"]}
 ignore = [
   "S101",  # asserts allowed
-  "INP001", # no implicit namespace
-  "D",  # ignore documentation for now
-  "D203",  # `one-blank-line-before-class` (D203) and `no-blank-line-before-class` (D211) are incompatible
-  "D212",  # `multi-line-summary-first-line` (D212) and `multi-line-summary-second-line` (D213) are incompatible
```
Borda, line 78: fixing these, so they do not need to be skipped

gaborbernat (approved):

PR #296, opened 2023-10-25
PackageDAG: Allow supressing the check for non-existent includes

tbabej:
At @proteinqure we use the `PackageDAG` to analyze the relationship between components in our stack but a recent addition to the `filter_nodes` breaks our workflows. Now I realize this is not part of a public API, but the ability to filter for only possibly present packages might be useful to others.

----

Adds an optional flag to suppress the check for non-existent includes. The flag defaults to True and hence the preserves the original behavior by default.

This is useful for filtering the DAG nodes by multiple packages, expected only some of them to match.

### DIFF
--- a/src/pipdeptree/_models/dag.py
+++ b/src/pipdeptree/_models/dag.py
@@ -89,7 +89,9 @@ class PackageDAG(Mapping[DistPackage, List[ReqPackage]]):
         node = self.get_node_as_parent(node_key)
         return self._obj[node] if node else []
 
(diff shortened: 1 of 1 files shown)


kemzeb:
If the maintainers don't mind, one approach to make this available to users is to introduce this as a command-line argument (or flag??), maybe something like:
```
pipdeptree --package-ignore-match-failure ab,cd,efg*
```

If it's going to be implemented as an argument make it mutually exclusive to the `-p` and `--packages` arguments.

gaborbernat (changes requested):
CI fails.

gaborbernat:
Stalled.

PR #302, opened 2023-11-14
Expect hpy in pypy versions >= 7.3.3

parona-source:
https://doc.pypy.org/en/latest/whatsnew-pypy3-7.3.2.html
https://doc.pypy.org/en/latest/whatsnew-pypy3-7.3.3.html

### DIFF
--- a/tests/test_non_host.py
+++ b/tests/test_non_host.py
@@ -33,7 +33,11 @@ def test_custom_interpreter(
     if implementation == "CPython":
         expected = {"pip", "setuptools", "wheel"}
     elif implementation == "PyPy":
-        expected = {"cffi", "greenlet", "pip", "readline", "setuptools", "wheel"}
+        # hpy added in 7.3.2, enabled in 7.3.3
+        if sys.pypy_version_info >= (7, 3, 3):  # type: ignore[attr-defined]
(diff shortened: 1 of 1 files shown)


gaborbernat (changes requested):
CI failing.

parona-source:
>  Failure. Coverage is below 100%.
 
 Covering this isn't possible without adding pypy3 into the ci. So would this be fine?

parona-source:
Went with disabling the coverage in the pypy phase.

Making coverage happy would require 1. adding pypy to the ci 2. splitting diff-cover into its own phase which depends on the python envs to finish.

Problem being that I haven't seen away in tox to glob all the reports after without explicitly writing out the envs again https://github.com/Bachmann1234/diff_cover#multiple-xml-coverage-reports

gaborbernat (changes requested):
Let's add pypy to the CI.

gaborbernat:
Seems this stalled, if you pick it up again we can reopen.

parona-source:
Came back to this whilst addressing on pypy failing in a different.

In this case the bundled hpy added setuptools to its Requires-Dist leading to the pretty printed pipdeptree output to be unreliable, decided to use the json output to have better output to work with.

For coverage decided to add a new dependency coverage_conditional_plugin because grafting the coverage data from several python implementations was arduous.

PR #303, opened 2023-11-23
models/package: have `version_spec` only exist in `ReqPackage`

kemzeb:
This behavior is specific to `ReqPackage`, so let's just have it exist in this class rather than having it in the `Package` parent class.

Just to be extra careful, I ran the rendering options that may be affected by this change (so graphviz, mermaid, and json tree) in both reversed and non-reversed order to cover the code paths that use `version_spec`. Everything looks fine from what I can see.


kemzeb:
Recent change includes 2 test cases to satisfy code coverage

gaborbernat (approved):

PR #325, opened 2024-03-02
Fix key mismatch between requirements and dists

### LINKED ISSUE #324: Issue with packages built with pdm-backend 2.x
Hi tox-dev team, I've filed a bug to setuptools but it is related to pipdeptree (manifests through it), I think you may want to be aware - https://github.com/pypa/setuptools/issues/4250.
TL;DR is packages built with `pdm-backend` 2.x use unexpected normalization for `.dist-info` folder name, causing pipdeptree to fail mapping requrement to dist.

f3flight:
It was found that keys are not pep503 normalized in `pkg_resources`, instead using different regex which retains dots, creating potential mismatch between keys in `Requirement` objects and keys in `DistInfoDistribution` objects.
This change fixes that by applying pep503 normalization on the keys we get from `pkg_resources` objects.

Additionally, it was found that `DistInfoDistribution` uses derived value for `project_name` based on `.dist-info` folder name, which may result in an imprecise package name (not the same as in metadata due to double normalization).
This change recovers true package name from metadata when possible.

Fixes #324

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 3.0.0
+## 2.7.0
 
 - Normalize keys with pep503 regex and recover package names from metadata instead of relying on DistInfoDistribution
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.0.0
+
+- Normalize keys with pep503 regex and recover package names from metadata instead of relying on DistInfoDistribution
+
 ## 2.6.0
 
 - Handle mermaid output for a reversed tree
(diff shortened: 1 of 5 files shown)


CHANGES.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.0.0
```
gaborbernat, line 3: I'd make it a bugfix version personally.

f3flight: I am worried that key changes my break somebody, basically it's not just a fix in matching but keys with dots will now be with dashes, so if someone relied on this they will have to make changes on their side. Hence major bump, but up to you.

gaborbernat: Any bug fix is a breaking change for someone. Let's make it to minor I'm fine breaking some probably a small percentage of users.

f3flight: Changed to 2.7.0, or would you prefer 2.6.1? @gaborbernat

gaborbernat: Is fine as is.

PR #337, opened 2024-03-23
Use GitHub issue forms

kemzeb:
This is so we can have more structured way of getting necessary information, automating tagging issues, among other useful features.

See https://github.com/kemzeb/pipdeptree/issues and press "New Issues" to see it in action.


kemzeb:
Making it a draft for now so that we can check for any typos or missing form data that should be added.

kemzeb:
Hey @gaborbernat, it looks like I might not have merge permissions? Though the "merging can be performed automatically with 0 approving reviews" message sounds odd.

gaborbernat:
@kemzeb you should be good now 👍

kemzeb:
Sounds good, thanks!

PR #340, opened 2024-04-02
Allow pip>=20.2.4

austinpray-mixpanel:
I was actively using pipdeptree to revive a legacy project stuck on 20.2.4 (https://pip.pypa.io/en/stable/user_guide/#changes-to-the-pip-dependency-resolver-in-20-3-2020). Requiring a more modern pip version breaks this use case. Furthermore could cause issues debugging docker images with pinned pip versions etc.

Is there any reason to make the pip dependency range so tight? This would allow people on pinned pip versions to install pipdeptree without it upsetting their environment.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ dynamic = [
   "version",
 ]
 dependencies = [
-  "pip>=23.1.2",
+  "pip>=20.2.4",
 ]
 optional-dependencies.graphviz = [
   "graphviz>=0.20.1",


gaborbernat:
There's a very simple reason for this. Inside the continuous integration system we only test for the latest version. Support for older versions are not guaranteed to work, and my break without us knowing about it without having a separate test environment specifically aimed at that. Given that if we going to relax this version you will need to add separate builds testing the new lower version, this is something that can be achieved by using tox-uv nowadays.

austinpray-mixpanel:
>you will need to add separate builds testing the new lower version, this is something that can be achieved by using tox-uv nowadays.

Yep happy to do this!

gaborbernat:
Closing as seems to have stalled.

Issue #343, opened 2024-04-05
`test_custom_interpreter` fails


mtelka:
### What pipdeptree version are you using?

2.17.0

### Are you running pipdeptree in a virtual environment?

- [ ] Yes

### Describe the problem

With version `2.17.0` the `test_custom_interpreter` test started to fail:
```
============================= test session starts ==============================
platform sunos5 -- Python 3.9.19, pytest-8.1.1, pluggy-1.4.0 -- $(BUILD_DIR)/.tox/py39/bin/python
cachedir: .tox/py39/.pytest_cache
rootdir: $(BUILD_DIR)
configfile: pyproject.toml
plugins: mock-3.14.0, cov-5.0.0
collecting ... collected 2 items

tests/test_non_host.py::test_custom_interpreter[True] FAILED             [ 50%]
tests/test_non_host.py::test_custom_interpreter[False] FAILED            [100%]

=================================== FAILURES ===================================
________________________ test_custom_interpreter[True] _________________________

tmp_path = PosixPath('/tmp/pytest-of-marcel/pytest-137/test_custom_interpreter_True_0')
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7fffacde7130>
capfd = <_pytest.capture.CaptureFixture object at 0x7fffacde7310>
args_joined = True

    @pytest.mark.parametrize("args_joined", [True, False])
    def test_custom_interpreter(
        tmp_path: Path,
        monkeypatch: pytest.MonkeyPatch,
        capfd: pytest.CaptureFixture[str],
        args_joined: bool,
    ) -> None:
        result = virtualenv.cli_run([str(tmp_path / "venv"), "--activators", ""])
        cmd = [sys.executable]
        monkeypatch.chdir(tmp_path)
        py = str(result.creator.exe.relative_to(tmp_path))
        cmd += [f"--python={result.creator.exe}"] if args_joined else ["--python", py]
        monkeypatch.setattr(sys, "argv", cmd)
        main()
        out, _ = capfd.readouterr()
        found = {i.split("==")[0] for i in out.splitlines()}
        implementation = python_implementation()
        if implementation == "CPython":
            expected = {"pip", "setuptools", "wheel"}
        elif implementation == "PyPy":
            expected = {"cffi", "greenlet", "pip", "readline", "setuptools", "wheel"}  # pragma: no cover
        else:
            raise ValueError(implementation)
        if sys.version_info >= (3, 12):
[...]

PR #346, opened 2024-04-06
refactor the non_host part for not injecting to custom env

xiacunshun:
fix: https://github.com/tox-dev/pipdeptree/issues/343

We can get the sitepackages directory from the given interpreter directly instead of running our code again with it.


xiacunshun:
I think the result of this method is the same with the original one. Please take a look if you have time. @kemzeb

tests/test_non_host.py
```diff
@@ -39,11 +39,3 @@ def test_custom_interpreter(
     if sys.version_info >= (3, 12):
         expected -= {"setuptools", "wheel"}
     assert found == expected, out
-
-    monkeypatch.setattr(sys, "argv", [*cmd, "--graph-output", "something"])
```
xiacunshun, line 43: We now can use --graph-output too, so I think we don't need this check anymore.

gaborbernat (dismissed):

src/pipdeptree/_discovery.py
```diff
@@ -1,26 +1,35 @@
...
     original_dists: Iterable[Distribution] = []
+    py_path = Path(interpreter).absolute()
 
     if local_only and in_venv:
         venv_site_packages = site.getsitepackages([sys.prefix])
         original_dists = distributions(path=venv_site_packages)
     elif user_only:
         original_dists = distributions(path=[site.getusersitepackages()])
+    elif py_path != Path(sys.executable).absolute():
+        cmd = [str(py_path), "-c", "import site; print(site.getsitepackages())"]
```
kemzeb, line 30: Would it make sense to grab the potential package metadata from `sys.path`  instead? It looks like `importlib.metadata` uses it by [default](https://github.com/python/cpython/blob/f5bd65ed37bbccfa113438cf9e34feae6816ecb1/Lib/importlib/metadata.py#L366) if you never supply a `path` argument to `distributions()`.

I fetched your branch and I do get the expected results.

tests/test_non_host.py
kemzeb: Yes, since that test snippet covers the now outdated graphviz check, and all we're doing now is running the host python but with the custom python paths it makes sense to remove it.

src/pipdeptree/_discovery.py
xiacunshun: I used `sitepackages` to keep the same as the code above. But you are right, it is more standard to use `sys` which `importlib.metadata` uses.

kemzeb: Yes, and now that I've been thinking about it further, we have been using the `sys.path` provided by the custom Python environment before these changes (i.e. when doing just `pipdeptree --python path/to/python`). So I think we should do this 👍

kemzeb (commented):
Hmm, I noticed that when we're using a venv that has system site packages enabled (i.e. I did `python -m venv --system-site-packages venv`), running the following:

```
pipdeptree --python path/to/python --local-only
```

...would cause us to hit the else branch (since I am not running in a virtual environment). We don't want this since it will give metadata paths from the entire environment.

I think what we need to do here is if --python and --local are given, we should grab the result of:
```
"str(py_path), "-c", "import sys, site; print(site.getsitepackages([sys.prefix]))"
```

...and pass it to `distributions()`. Otherwise, use `sys.path`.

src/pipdeptree/_discovery.py
```diff
@@ -1,26 +1,37 @@
...
     in_venv = sys.prefix != sys.base_prefix
     original_dists: Iterable[Distribution] = []
+    py_path = Path(interpreter).absolute()
 
     if local_only and in_venv:
         venv_site_packages = site.getsitepackages([sys.prefix])
         original_dists = distributions(path=venv_site_packages)
     elif user_only:
         original_dists = distributions(path=[site.getusersitepackages()])
+    elif py_path != Path(sys.executable).absolute():
```
kemzeb, line 29: Building on top of my previous comment, I think we could do something similar to the following:
```suggestion
   using_custom_interpreter = py_path != Path(sys.executable).absolute()
   if using_custom_interpreter:
       args = []
       if local_only:
           args = ["str(py_path), "-c", "import sys, site; print(site.getsitepackages([sys.prefix]))"]
       else:
           cmd = [str(py_path), "-c", "import sys, site; print(site.getsitepackages([sys.prefix]))"]
        # Add the custom python subprocess run code here
    if local_only and in_venv:
        venv_site_packages = site.getsitepackages([sys.prefix])
        original_dists = distributions(path=venv_site_packages)
    elif user_only:
        original_dists = distributions(path=[site.getusersitepackages()])
```

xiacunshun:
> Hmm, I noticed that when we're using a venv that has system site packages enabled (i.e. I did `python -m venv --system-site-packages venv`), running the following:
> 
> ```
> pipdeptree --python path/to/python --local-only
> ```
> 
> ...would cause us to hit the else branch (since I am not running in a virtual environment). We don't want this since we will use paths to metadata from the entire environment.
> 
> I think what we need to do here is if --python and --local are given, we should grab the result of:
> 
> ```
> "str(py_path), "-c", "import sys, site; print(site.getsitepackages([sys.prefix]))"
> ```
> 
> ...and pass it to `distributions()`. Otherwise, use `sys.path`.

That's a really good situation which I didn't notice. I will add some tests to cover this too.

tests/test_non_host.py
```diff
@@ -40,10 +40,32 @@ def test_custom_interpreter(
...
-    assert err == "graphviz functionality is not supported when querying non-host python\n"
+
+def test_custom_interpreter_with_local_only(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+    capfd: pytest.CaptureFixture[str],
+) -> None:
+    venv_path = str(tmp_path / "venv")
+
+    result = virtualenv.cli_run([venv_path, "--system-site-packages", "--activators", ""])
```
xiacunshun, line 51: When I was adding tests, I checked the [code](https://github.com/pypa/virtualenv/blob/main/src/virtualenv/run/__init__.py#L20) and [manual](https://virtualenv.pypa.io/en/latest/user_guide.html#programmatic-api) of `virtualenv.cli_run`and it looks like it has `options` and `env` args which we can pass. However, it seems adding args to cmd is much simpler, so let's directly use this way.

xiacunshun: I originally wanted to construct a `foo` dist in other directory then pending it to `sys.path`, but I noticed that `system-site-packages` is enough because there are at least two dists(`packaging` and `pipdeptree`) in system dists than in venv.

tests/test_non_host.py
```diff
@@ -40,10 +40,32 @@ def test_custom_interpreter(
...
+
+    cmd = [sys.executable]
+    monkeypatch.chdir(tmp_path)
+    cmd += [f"--python={result.creator.exe}"]
+    cmd += ["--local-only"]
```
kemzeb, line 56: Let's remove the chdir monkeypatch since it doesn't look like it affects the test in any way.

Let's also combine `cmd` elements into one line:

```suggestion
    cmd = [sys.executable, f"--python={result.creator.exe}", "--local-only"]
```

Also, it doesn't actually matter what's in the first element since we are just parsing these arguments. Instead of `sys.exectuable`, let's add just the empty string for clarity.

src/pipdeptree/_discovery.py
```diff
@@ -1,26 +1,45 @@
...
+        # We query the interpreter directly to get the `path` which is used by distributions(),
+        # so we can search the customer dists with it.
+        # If --python and --local-only are given at the same time, it means the dists only in
+        # customer's env are needed. (.e.g --system-site-packages is given when creating venv)
+        if local_only:
+            args = "import sys, site; print(site.getsitepackages([sys.prefix]))"
+        else:
+            args = "import sys; print(sys.path)"
+
+        cmd = [str(py_path), "-c", args]
```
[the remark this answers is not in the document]
xiacunshun: done

tests/test_non_host.py
xiacunshun: done

src/pipdeptree/_discovery.py
```diff
@@ -1,26 +1,45 @@
...
+    using_custom_interpreter = py_path != Path(sys.executable).absolute()
 
-    if local_only and in_venv:
+    if user_only:
+        original_dists = distributions(path=[site.getusersitepackages()])
+    elif using_custom_interpreter:
+        # We query the interpreter directly to get the `path` which is used by distributions(),
+        # so we can search the customer dists with it.
+        # If --python and --local-only are given at the same time, it means the dists only in
+        # customer's env are needed. (.e.g --system-site-packages is given when creating venv)
```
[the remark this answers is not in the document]
xiacunshun: done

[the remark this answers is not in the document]
xiacunshun: I'm sorry, I seem to have misunderstood. I thought it was sufficient to ensure the length shorter than 120 characters.

src/pipdeptree/_discovery.py
```diff
@@ -1,26 +1,44 @@
...
 
-    if local_only and in_venv:
+    if user_only:
+        original_dists = distributions(path=[site.getusersitepackages()])
+    elif using_custom_interpreter:
+        # We query the interpreter directly to get its `sys.path` list to be used by `distributions()`.
+        # If --python and --local-only are given, we ensure that we are only using paths associated to the interpreter's
+        # environment.
+        if local_only:
+            cmd = "import sys, site; print(site.getsitepackages([sys.prefix]))"
```
kemzeb, line 32: Now that I think about it, it looks like that their is a potential to miss out on metadata when dealing with a custom interpreter or a virtual environment if we just use `site.getsitepackages()` . I think it makes sense to filter `sys.path` for environment-specific packages as most of the time `sys.path` already holds these site-packages directories.

We can filter `sys.path` by doing something like the following:
```suggestion
            cmd = "import sys; print([p for p in sys.path if p.startswith(sys.prefix)])"
```

Though it would make this PR not atomic, we should also do this in the `elif local_only and in_venv` branch.

`pip list` also does this by default when users pass `--local` and they're in a virtual env; see [here](https://github.com/pypa/pip/blob/31437d675abfa1563cffb3f68a0a23f3cbd68afa/src/pip/_internal/metadata/importlib/_envs.py#L170) (uses sys.path by default) and [here](https://github.com/pypa/pip/blob/31437d675abfa1563cffb3f68a0a23f3cbd68afa/src/pip/_internal/utils/misc.py#L368) (this function is used by `BaseDistribution.local`).

kemzeb:
FYI once we merge this I'll make a 2.18.0 release.

src/pipdeptree/_discovery.py
xiacunshun: It makes sense, done.

xiacunshun: Tests also need to adjust since we get path to search directly from `sys.path`.
Done this.

tests/test_non_host.py
```diff
@@ -40,10 +40,29 @@ def test_custom_interpreter(
...
+    monkeypatch: pytest.MonkeyPatch,
+    capfd: pytest.CaptureFixture[str],
+) -> None:
+    venv_path = str(tmp_path / "venv")
+
+    result = virtualenv.cli_run([venv_path, "--system-site-packages", "--activators", ""])
+
+    cmd = ["", f"--python={result.creator.exe}", "--local-only"]
+    monkeypatch.setattr(sys, "prefix", venv_path)
+    monkeypatch.setattr(sys, "argv", cmd)
```
kemzeb, line 55: Turns out monkey patching is dangerous to do as it actually changes `sys.prefix` and `sys.argv` directly. Let's use `pytest_mock.MockerFixture` instead and mock where they are "looked up" instead (see [here](https://docs.python.org/3/library/unittest.mock.html#id6) for more details). Like this:
```suggestion
   mocker: MockerFixture,
    capfd: pytest.CaptureFixture[str],
) -> None:
    venv_path = str(tmp_path / "venv")

    result = virtualenv.cli_run([venv_path, "--system-site-packages", "--activators", ""])

    cmd = ["", f"--python={result.creator.exe}", "--local-only"]
    mocker.patch("pipdeptree._discovery.sys.prefix", venv_path)
    mocker.patch("pipdeptree._discovery.sys.argv", cmd)
```

tests/_models/test_package.py
```diff
@@ -26,11 +27,13 @@ def test_guess_version_setuptools(mocker: MockerFixture) -> None:
...
 
 
-def test_package_as_frozen_repr(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
+def test_package_as_frozen_repr(tmp_path: Path, mocker: MockerFixture) -> None:
     file_path = tmp_path / "foo.egg-link"
     with Path(file_path).open("w") as f:
         f.write("/A/B/foo")
-    monkeypatch.syspath_prepend(str(tmp_path))
+    mock_path = sys.path.copy()
+    mock_path.append(str(tmp_path))
```
xiacunshun, line 35: We need to keep the original part of `sys.path` because `as_frozen_repr` API need them.

tests/test_non_host.py
xiacunshun: Done all.

kemzeb: I appreciate you looking for other areas where monkeypatch was being misused, but next time please just change the requested areas so that the PR isn't trying to do too many things at once; another PR could be opened for the monkeypatch misuse 👍

tests/test_discovery.py
```diff
@@ -27,8 +27,9 @@ def test_local_only(tmp_path: Path, monkeypatch: pytest.MonkeyPatch, capfd: pyte
         f.write("Metadata-Version: 2.3\n" "Name: foo\n" "Version: 1.2.5\n")
 
     cmd = [str(result.creator.exe.parent / "python3"), "--local-only"]
-    monkeypatch.setattr(sys, "prefix", venv_path)
-    monkeypatch.setattr(sys, "argv", cmd)
+    mocker.patch("pipdeptree._discovery.sys.prefix", venv_path)
+    mocker.patch("pipdeptree._discovery.sys.path", venv_site_packages)
```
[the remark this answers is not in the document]
xiacunshun: fixed.

kemzeb:
Can't see any other issues, so it looks like this is good to go. Will release 2.18.0 after merging

kemzeb (approved):

src/pipdeptree/_discovery.py
```diff
@@ -1,26 +1,44 @@
...
+        # We query the interpreter directly to get its `sys.path` list to be used by `distributions()`.
+        # If --python and --local-only are given, we ensure that we are only using paths associated to the interpreter's
+        # environment.
+        if local_only:
+            cmd = "import sys; print([p for p in sys.path if p.startswith(sys.prefix)])"
+        else:
+            cmd = "import sys; print(sys.path)"
+
+        args = [str(py_path), "-c", cmd]
+        result = subprocess.run(args, stdout=subprocess.PIPE, check=False)  # noqa: S603
```
gaborbernat, line 37: text=True and no need for stdout decode

gaborbernat (changes requested):

src/pipdeptree/_discovery.py
xiacunshun: fixed.

kemzeb:
Looks like we're good to go, releasing 2.18.0

Issue #348, opened 2024-04-09
2.17.0: pytest fails in 3 units


kloczek:
### What pipdeptree version are you using?

2.17.0

### Are you running pipdeptree in a virtual environment?

- [ ] Yes

### Describe the problem

I'm packaging your module as an rpm package so I'm using the typical PEP517 based build, install and test cycle used on building packages from non-root account.
- `python3 -sBm build -w --no-isolation`
- because I'm calling `build` with `--no-isolation` I'm using during all processes only locally installed modules
- install .whl file in </install/prefix> using `installer` module
- run pytest with $PYTHONPATH pointing to sitearch and sitelib inside </install/prefix>
- build is performed in env which is *`cut off from access to the public network`* (pytest is executed with `-m "not network"`)

I'm not sure why one of those units are trying to execute /usr/bin/pipdeptree if pipdeptree is installed in $PYTHONPATH prefix.
<details>
<summary>Here is pytest output:</summary>

```console
+ PYTHONPATH=/home/tkloczko/rpmbuild/BUILDROOT/python-pipdeptree-2.17.0-2.fc37.x86_64/usr/lib64/python3.10/site-packages:/home/tkloczko/rpmbuild/BUILDROOT/python-pipdeptree-2.17.0-2.fc37.x86_64/usr/lib/python3.10/site-packages
+ /usr/bin/pytest -ra -m 'not network'
==================================================================================== test session starts ====================================================================================
platform linux -- Python 3.10.14, pytest-8.1.1, pluggy-1.4.0
rootdir: /home/tkloczko/rpmbuild/BUILD/pipdeptree-2.17.0
configfile: pyproject.toml
plugins: mock-3.14.0
collected 96 items

tests/_models/test_dag.py ...........                                                                                                                                                 [ 11%]
tests/_models/test_package.py ....................                                                                                                                                    [ 32%]
tests/render/test_graphviz.py ...                                                                                                                                                     [ 35%]
tests/render/test_json_tree.py ..                                                                                                                                                     [ 37%]
tests/render/test_mermaid.py ..                                                                                                                                                       [ 39%]
tests/render/test_render.py .....                                                                                                                                                     [ 44%]
tests/render/test_text.py ....................                                                                                                                                        [ 65%]
tests/test_cli.py .................                                                                                                                                                   [ 83%]
tests/test_discovery.py ...                                                                                                                                                           [ 86%]
tests/test_non_host.py FF                                                                                                                                                             [ 88%]
[...]

PR #351, opened 2024-04-10
Fix test_console failure finding pipdeptree in other environments

xiacunshun:
When we are using pytest in packaging process, we may not get the `/usr/bin/pipdeptree` as we need. So, let's use the `PATH` env to get the binary instead.

fix: https://github.com/tox-dev/pipdeptree/issues/348

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,7 @@ optional-dependencies.test = [
   "covdefaults>=2.3",
   "diff-cover>=8.0.1",
   "pytest>=7.4.3",
+  "pytest-console-scripts>=1.4.1",
   "pytest-cov>=4.1",
   "pytest-mock>=3.12",
   "virtualenv<21,>=20.25",
--- a/tests/test_pipdeptree.py
+++ b/tests/test_pipdeptree.py
@@ -1,13 +1,17 @@
 from __future__ import annotations
 
 import sys
-from pathlib import Path
 from subprocess import check_call  # noqa: S404
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from pytest_console_scripts import ScriptRunner
 
 
 def test_main() -> None:
     check_call([sys.executable, "-m", "pipdeptree", "--help"])
 
 
-def test_console() -> None:
-    check_call([Path(sys.executable).parent / "pipdeptree", "--help"])
+def test_console(script_runner: ScriptRunner) -> None:
+    result = script_runner.run("pipdeptree", "--help")
+    assert result.success


gaborbernat (changes requested):
I don't like this, if someone's has another pipdeptree on an earlier path we'd wrongfully accept it.

kemzeb:
Since it looks like this test exists to ensure that our `pipdeptree` console script exists, I believe an alternative way to test this without relying on a file path is to utilize `importlib.metdata.Distribution.entry_points()`. We can do something like the following:
```python
d = distribution("pipdeptree")
our_entry_points = d.entry_points
assert len(our_entry_points) == 1
pipdeptree_entry_point = our_entry_points[0]
pipdeptree_module = pipdeptree_entry_point.load()
return_code = pipdeptree_module()
assert return_code == 0
```

The `load()` call will return the module associated to the entry point (so `pipdeptree.__main__`) and we can execute it from there.

tests/test_pipdeptree.py
```diff
@@ -1,13 +1,20 @@
...
 
-def test_console() -> None:
-    check_call([Path(sys.executable).parent / "pipdeptree", "--help"])
+def test_console(mocker: MockerFixture) -> None:
+    d = Distribution.from_name("pipdeptree")
+    our_entry_points = d.entry_points
+    assert len(our_entry_points) == 1
+    pipdeptree_entry_point = next(iter(our_entry_points))
+    pipdeptree_module = pipdeptree_entry_point.load()
+    mocker.patch("pipdeptree.__main__.sys.argv", [""])
```
xiacunshun, line 18: It looks args will be passed to pipdeptree if not empty them.
```
message = '__main__.py: error: unrecognized arguments: --color=yes --cov /data/upstream/cunshunxia/pipdeptree/.tox/py311/lib/pyt...hunxia/pipdeptree/.tox/coverage.py311.xml --junitxml /data/upstream/cunshunxia/pipdeptree/.tox/junit.py311.xml tests\n'
```

kemzeb: Instead of mocking, just do `pipdeptree_module([])` since `load()` is actually having us call our `main()` when it parses `pipdeptree.__main__:main` (`argparse` defaults to using `sys.argv` if we don't pass it anything).

gaborbernat (dismissed):

tests/test_pipdeptree.py
xiacunshun: Great. Done this.

kemzeb (dismissed):

xiacunshun:
Emm, sorry, I retry this test in packaging, but looks failed also?
```
____________________________________________________ test_console ____________________________________________________

    def test_console() -> None:
        d = Distribution.from_name("pipdeptree")
        our_entry_points = d.entry_points
>       assert len(our_entry_points) == 1
E       assert 0 == 1
E        +  where 0 = len([])
```

kemzeb:
> Emm, sorry, I retry this test in packaging, but looks failed also?
> 
> ```
> ____________________________________________________ test_console ____________________________________________________
> 
>     def test_console() -> None:
>         d = Distribution.from_name("pipdeptree")
>         our_entry_points = d.entry_points
> >       assert len(our_entry_points) == 1
> E       assert 0 == 1
> E        +  where 0 = len([])
> ```

That's weird. It's able to find pipdeptree's metadata but can't find its console scripts? Can you provide the steps you ran to run the test

xiacunshun:
I think I figure it out....
After the step "%generate_buildrequires" in spec file, a dist-info is created in source directory.
```
<mock-chroot> sh-5.2# ls pipdeptree-2.18.0/
LICENSE  README.md  pipdeptree-2.18.0.dist-info  pyproject.toml  src  tests  tox.ini
```
But it only has a METADATA under it.
```
<mock-chroot> sh-5.2# ls pipdeptree-2.18.0/pipdeptree-2.18.0.dist-info/
METADATA
```
When we run pytest, current directory is added to `sys.path` which cause `distribution` can find the fake dist-info and recognizes it as the `pipdeptree`.

xiacunshun:
I am not sure why they do this, but it looks affect our testcases.
see https://src.fedoraproject.org/rpms/python-pipdeptree/blob/rawhide/f/python-pipdeptree.spec

xiacunshun:
> > Emm, sorry, I retry this test in packaging, but looks failed also?
> > ```
> > ____________________________________________________ test_console ____________________________________________________
> > 
> >     def test_console() -> None:
> >         d = Distribution.from_name("pipdeptree")
> >         our_entry_points = d.entry_points
> > >       assert len(our_entry_points) == 1
> > E       assert 0 == 1
> > E        +  where 0 = len([])
> > ```
> 
> That's weird. It's able to find pipdeptree's metadata but can't find its console scripts? Can you provide the steps you ran before running the test

I download their src.rpm([here](https://kojipkgs.fedoraproject.org//packages/python-pipdeptree/2.17.0/1.fc41/src/python-pipdeptree-2.17.0-1.fc41.src.rpm)), uncompress it, replace the sources with our new code and then rpmbuild -ba ~/rpmbuild/SPECS/python-pipdeptree.spec.
(There are also some steps to deal with dependencies.)

kemzeb:
So from reading the Fedora [spec](https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_dist_info_metadata) dealing with Python packages and this Python [spec](https://packaging.python.org/en/latest/specifications/recording-installed-packages/#the-entry-points-txt-file), it looks like packaging tools aren't required to create an entry_points.txt metadata file if they don't want to. But they **must at least** generate a METADATA file.

I'm not familiar with the packaging process, let alone with Fedora, but this does reflect your outputs since we only see the METADATA file (although the Fedora docs explain that the source and the metadata should be installed in "%{python3_sitelib} or %{python3_sitearch}" rather than within the source directory?).

Not too sure why it is not required for entry_points.txt to be generated when there are console scripts that exist, but because of this testing by using the importlib API here may only work for package installers that do generate entry_points.txt if possible (e.g. pip).

kemzeb:
There looks to be a pytest plugin just for testing console scripts, and from looking at it source it seems that it will either:
- Load the script using `importlib.metadata.entry_points()` (see [here](https://github.com/kvas-it/pytest-console-scripts/blob/d873f69541560c35db8783ea86a413fb48226d4b/pytest_console_scripts/__init__.py#L298))
- Look for the script in either PATH or the cwd if it doesn't find the entry point using `importlib.metadata` (see [here](https://github.com/kvas-it/pytest-console-scripts/blob/d873f69541560c35db8783ea86a413fb48226d4b/pytest_console_scripts/__init__.py#L280))

xiacunshun:
> There looks to be a pytest plugin just for testing console scripts, and from looking at the source it seems that it will either:
> 
> * Load the script using `importlib.metadata.entry_points()` (see [here](https://github.com/kvas-it/pytest-console-scripts/blob/d873f69541560c35db8783ea86a413fb48226d4b/pytest_console_scripts/__init__.py#L298))
> * Look for the script in either PATH or the cwd if it doesn't find the entry point when using `importlib.metadata` (see [here](https://github.com/kvas-it/pytest-console-scripts/blob/d873f69541560c35db8783ea86a413fb48226d4b/pytest_console_scripts/__init__.py#L280))

This approach worked and looks solve the problem in packaging!
I retried, and console test passed.

pyproject.toml
```diff
@@ -50,6 +50,7 @@ optional-dependencies.test = [
   "covdefaults>=2.3",
   "diff-cover>=8.0.1",
   "pytest>=7.4.3",
+  "pytest-console-scripts",
```
gaborbernat, line 53: please lower bound it

gaborbernat, line 53: please lower bound it

gaborbernat (changes requested):

gaborbernat (approved):

pyproject.toml
xiacunshun: Done.

kemzeb (approved):

kemzeb:
Will release 2.18.1 after merging this

PR #356, opened 2024-04-19
Use dropdown instead of checkbox in bug report template

kemzeb:
Somehow missed the dropdown menu type when reading the [docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#dropdown) and to me this reads much better.

See an example by going to https://github.com/kemzeb/pipdeptree/issues and then navigate to `Report a Bug -> Get Started`.


kemzeb:
Gonna merge since this is a very low-impact and simple change.

PR #357, opened 2024-04-20
Have duplicate metadata and invalid req strings honor --warn option

### LINKED ISSUE #355: Consider --warn option when dealing with duplicate metadata and invalid requirement string warnings

2.18.1

- [ ] Yes

### Describe the problem

The new duplicate metadata warning text and invalid requirement string warning text (which I both added) do not account for the --warn option. An example:
```console
vscode ➜ /workspaces/pipdeptree (main) $ pipdeptree --warn silence
Warning!!! Duplicate package metadata found:
"/usr/local/lib/python3.8/site-packages"
  pip                              23.0.1           (using 24.0, "/home/vscode/.local/lib/python3.8/site-packages")
------------------------------------------------------------------------
covdefaults==2.3.0
└── coverage [required: >=6.0.2, installed: 7.4.4]
. . .
```

kemzeb:
This resolves #355 by making changes and refactors to the warning logic. It does so by introducing a module-level singleton "WarningPrinter" object and refactors the code in such a way to integrate this object for it to be used.

### DIFF
--- a/src/pipdeptree/_discovery.py
+++ b/src/pipdeptree/_discovery.py
@@ -10,6 +10,8 @@ from typing import Iterable, Tuple
 
 from packaging.utils import canonicalize_name
 
+from pipdeptree._warning import get_warning_printer
+
 
 def get_installed_distributions(
     interpreter: str = str(sys.executable),
@@ -42,6 +44,8 @@ def get_installed_distributions(
     else:
         original_dists = distributions()
 
+    warning_printer = get_warning_printer()
+
     # Since importlib.metadata.distributions() can return duplicate packages, we need to handle this. pip's approach is
     # to keep track of each package metadata it finds, and if it encounters one again it will simply just ignore it. We
     # take it one step further and warn the user that there are duplicate packages in their environment.
@@ -55,11 +59,17 @@ def get_installed_distributions(
             seen_dists[normalized_name] = dist
             dists.append(dist)
             continue
-        already_seen_dists = first_seen_to_already_seen_dists_dict.setdefault(seen_dists[normalized_name], [])
-        already_seen_dists.append(dist)
+        if warning_printer.should_warn():
+            already_seen_dists = first_seen_to_already_seen_dists_dict.setdefault(seen_dists[normalized_name], [])
+            already_seen_dists.append(dist)
 
-    if first_seen_to_already_seen_dists_dict:
-        render_duplicated_dist_metadata_text(first_seen_to_already_seen_dists_dict)
+    should_print_warning = warning_printer.should_warn() and first_seen_to_already_seen_dists_dict
+    if should_print_warning:
+        warning_printer.print_multi_line(
+            "Duplicate package metadata found",
+            lambda: render_duplicated_dist_metadata_text(first_seen_to_already_seen_dists_dict),
+            ignore_fail=True,
+        )
 
     return dists
 
@@ -77,7 +87,6 @@ def render_duplicated_dist_metadata_text(
             dist_list = entries_to_pairs_dict.setdefault(entry, [])
             dist_list.append((first_seen, dist))
 
-    print("Warning!!! Duplicate package metadata found:", file=sys.stderr)  # noqa: T201
     for entry, pairs in entries_to_pairs_dict.items():
         print(f'"{entry}"', file=sys.stderr)  # noqa: T201
         for first_seen, dist in pairs:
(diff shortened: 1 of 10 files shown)


kemzeb:
Still working on this, I'll see if I can finish this and open it for review by the end of today.

kemzeb:
Did a redesign and this seems to be a better implementation than the last. Just need to add coverage tests then I'll open this PR for review

kemzeb:
I think this is ready to go, going to open this for review

kemzeb:
Here are my design decisions:
- Decided to keep the warning logic in their respective modules since they appear to be tightly coupled
- Included a `_warning` module; did try to add it to `render` but I encountered dependency cycle errors
- Implemented a `WarningPrinter` class, where I also exposed a shared module-level singleton of it
- Decided that the duplicate metadata warning should not be considered a "fail" warning when `--warn fail` is passed. My reasoning for this is that Python allows these duplicate distributions to exist and pip actively ignores it.

src/pipdeptree/_discovery.py
```diff
@@ -55,11 +59,17 @@ def get_installed_distributions(
...
-        already_seen_dists = first_seen_to_already_seen_dists_dict.setdefault(seen_dists[normalized_name], [])
-        already_seen_dists.append(dist)
-
-    if first_seen_to_already_seen_dists_dict:
-        render_duplicated_dist_metadata_text(first_seen_to_already_seen_dists_dict)
+        if warning_printer.should_warn():
+            already_seen_dists = first_seen_to_already_seen_dists_dict.setdefault(seen_dists[normalized_name], [])
+            already_seen_dists.append(dist)
+
+    should_print_warning = warning_printer.should_warn() and first_seen_to_already_seen_dists_dict
```
xiacunshun, line 66: Like `dist_name_to_invalid_reqs_dict` in dag.py, it looks like we only need to check if `first_seen_to_already_seen_dists_dict` is empty if you prefer to keep the if statement.

src/pipdeptree/_warning.py
```diff
@@ -0,0 +1,72 @@
...
+
+_shared_warning_printer = WarningPrinter()
+
+
+def get_warning_printer() -> WarningPrinter:
+    """Shared warning printer, representing a module-level singleton object."""
+    return _shared_warning_printer
+
+
+def parse_warning_type(type_str: str) -> WarningType:
```
xiacunshun, line 61: Can we directly do this in `build_parser`?  We can use `type=str.upper` to keep the letter uppercase if needed.

kemzeb: I do like this idea as we can move all the parsing logic into where it is supposed to be: in the CLI parser. argparse doesn't really like working with enums though, so I had to implement a custom `argparse.Action` class.

src/pipdeptree/_models/dag.py
```diff
@@ -65,7 +63,8 @@ def from_pkgs(cls, pkgs: list[Distribution]) -> PackageDAG:
                     req = next(requires_iterator)
                 except InvalidRequirementError as err:
                     # We can't work with invalid requirement strings. Let's warn the user about them.
-                    dist_name_to_invalid_reqs_dict.setdefault(p.project_name, []).append(str(err))
+                    if warning_printer.should_warn():
```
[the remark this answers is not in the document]
kemzeb: My reasoning behind this was that users who pass --warn silence would be building these data structures that would end up never being used since warnings can't be printed. However, it also feels weird having to keep performing this check even though its return value should never change.

xiacunshun:
It seems there are no big problems, but there's one thing that looks weird. As you said, we consider "duplicated" not as a failure, but if we specify `-w fail`, it still appears (even though there's a prompt). So I think we just drop it? (only warn it when using suppress).

src/pipdeptree/_models/dag.py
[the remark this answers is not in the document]
xiacunshun: You're right, if we don't check here, this part of the data structures would be wasted and there would be unnecessary time consumption. It seems better to keep the current situation.

xiacunshun:
Ignore my previous comment, I've reconsidered and the handling of "duplicated" here should be fine.

xiacunshun:
No questions for me here. But I don't have the approve auth, so maybe need to wait for @gaborbernat to take a further look.

gaborbernat (approved):

kemzeb:
Great, will release 2.19.0 after merging this

PR #359, opened 2024-04-29
Point "Changelog" project URL at Releases page

jwodder:


pyproject.toml
```diff
@@ -55,7 +55,6 @@ optional-dependencies.test = [
   "pytest-mock>=3.12",
   "virtualenv<21,>=20.25",
 ]
-urls.Changelog = "https://github.com/tox-dev/pipdeptree/blob/main/CHANGES.md"
```
gaborbernat, line 58: You should point to the GitHub releases page instead.

gaborbernat (changes requested):

pyproject.toml
jwodder: Done.

gaborbernat (approved):

PR #364, opened 2024-05-07
Mimic pip when handling --user-only in a virtual environment

kemzeb:
In v2.16.1, we would not print any user-site packages if we were:
- In a virtual environment
- Using a custom interpreter

This makes since as we should be isolated away from the system environment. The only case where we should be seeing user-site packages is when these environments have system site packages enabled. This patch brings back this behavior.


gaborbernat (approved):

kemzeb:
Will release 2.20.0 after successful merge and workflows

PR #366, opened 2024-05-15
Implement --python auto for environment detection

### LINKED ISSUE #362: Respect the active virtual environment instead of the installed location
### Describe the feature

If `pipdeptree` is installed in the system python location (to make it always accessible), it will by default use the system python to look for packages. When a virtual environment is activated, this causes a confusing situation where packages listed do not correspond to the activated environment.

To work around this, a user has to manually provide the desired python:
```shell
# activate env
conda activate <env>  # or other virtualenv equivalent
# look for package
pipdeptree -p <package> --local-only --python "$(which python)"
```

Installing `pipdeptree` as a system-wide utility seems relevant to avoid reinstalling it each time for each environment (similar to what `poetry` recommends). 

It seems to me that `pipdeptree` should prioritize the "active" python rather than its installed location.

kemzeb:
Resolves #362.

### DIFF
--- a/src/pipdeptree/_detect_env.py
+++ b/src/pipdeptree/_detect_env.py
@@ -1,0 +1,94 @@
+from __future__ import annotations
+
+import os
+import platform
+import subprocess  # noqa: S404
+import sys
+from pathlib import Path
+from typing import Callable
+
+
+def detect_active_interpreter() -> str:
+    """
+    Attempt to detect a venv, virtualenv, poetry, or conda environment by looking for certain markers.
+
+    If it fails to find any, it will fallback to using the system interpreter.
+    """
+    detection_funcs: tuple[Callable[[], Path | None]] = (
+        detect_venv_or_virtualenv_interpreter,
+        detect_conda_env_interpreter,
+        detect_poetry_env_interpreter,
+    )
+
+    for detect in detection_funcs:
+        path = detect()
+        if not path:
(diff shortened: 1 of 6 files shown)


kemzeb:
Manually tested using venv, virtualenv, conda, and poetry environments and they appear fine. I also virtualenv and conda using PyPy and they also appear fine (not sure if venv can be used with pypy and poetry makes it easy for us by giving us the interpreter directly so I don't believe we have to worry about it).

Making this a draft as I want to test this on a Windows machine.

src/pipdeptree/_detect_env.py
```diff
@@ -0,0 +1,94 @@
...
+
+    for detect in detection_funcs:
+        path = detect()
+        if not path:
+            continue
+        if not path.exists():
+            break
+        return str(path)
+
+    return sys.executable
```
kemzeb, line 31: I'm wondering if it makes sense to fallback to the system interpreter if we fail to detect any virtual environments?

kemzeb:
Tested these environments in a Windows 10 VM and they all appear to work fine.

The next thing I'm going to do is make auto-detection fail instead of falling back to the interpreter that pipdeptree is being ran with. To me, passing anything to `--python` implies that we should run an interpreter other than the one that we are using.

gaborbernat (approved):

PR #377, opened 2024-06-07
Use non-deprecated pytest-console-scripts API

kemzeb:
This removes a deprecation warning message that appears when running the automated tests.


gaborbernat (approved):

kemzeb:
Not sure why the first workflow has been in "Queue" status for a long time; I'm not able to cancel it either 🤔. I don't think it should be a problem

Issue #372, opened 2024-05-28
Report what virtual environment was chosen when using --python auto


kemzeb:
### Describe the feature

Referenced from https://github.com/tox-dev/pipdeptree/issues/362#issuecomment-2123099417 (adding it here so that we can track it):

> One thing that could be useful is to display the resolved environment when --python auto is used (or maybe even every time regardless?), just to validate the result is as expected. Something like a log message Resolved Python: [/abs/path/to/python] before the package tree listing would do the trick.

kemzeb:
Closed by #379.

PR #379, opened 2024-06-12
Print resolved interpreter when using env detection

stasfilin:
Good evening,

This is a super quick fix for #372

image

> One thing that could be useful is to display the resolved environment when --python auto is used (or maybe even every time regardless?)

What I did was create a new param, `log_resolved.` If True, print the resolved Python path.

I have only activated this to the `--python auto` option.


stasfilin:
@kemzeb Good evening,
Please review this MR when you have time. 

Thanks

kemzeb (changes requested):
This looks fine, just a couple of things.

src/pipdeptree/_detect_env.py
```diff
@@ -8,11 +8,16 @@
 from typing import Callable
 
 
-def detect_active_interpreter() -> str:
+def detect_active_interpreter(*, log_resolved: bool = False) -> str:
```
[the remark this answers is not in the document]
stasfilin: @kemzeb Yeah, you're right. Moved to main

src/pipdeptree/_detect_env.py
```diff
@@ -25,7 +30,10 @@ def detect_active_interpreter() -> str:
             continue
         if not path.exists():
             break
-        return str(path)
+        resolved_path = str(path)
+        if log_resolved:
+            print(f"Resolved Python: [{resolved_path}]")  # noqa: T201
```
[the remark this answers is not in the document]
stasfilin: @kemzeb fixed

tests/test_detect_env.py
```diff
@@ -52,3 +52,17 @@ def test_detect_active_interpreter_non_existent_path(
...
+) -> None:
+    mocker.patch("pipdeptree._detect_env.os.environ", {"VIRTUAL_ENV": str(tmp_path)})
+    mocker.patch("pipdeptree._detect_env.Path.exists", return_value=True)
+
+    actual_path = detect_active_interpreter()
+
+    captured = capsys.readouterr()
+
+    assert actual_path.startswith(str(tmp_path))
+    assert f"Resolved Python: [{tmp_path!s}" in captured.out
```
kemzeb, line 68: It looks like this should become an integration test since we now need to call `main()` instead of `detect_active_interpreter()`. I think we can just assert that the output `startswith()` the python interpreter message.

kemzeb (changes requested):

tests/test_detect_env.py
stasfilin: @kemzeb moved test to `test_main.py`. I mocked subprocess

src/pipdeptree/__main__.py
```diff
@@ -26,7 +26,9 @@ def main(args: Sequence[str] | None = None) -> None | int:
     warning_printer.warning_type = options.warn
 
     if options.python == "auto":
-        options.python = detect_active_interpreter()
+        resolved_path = detect_active_interpreter()
+        options.python = resolved_path
+        print(f"Resolved Python: [{resolved_path}]", file=sys.stderr)  # noqa: T201
```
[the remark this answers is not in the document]
stasfilin: Good point
Fixed

tests/test_main.py
```diff
@@ -0,0 +1,38 @@
...
+from pipdeptree.__main__ import main
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    import pytest
+    from pytest_mock import MockFixture
+
+
+def test_main_log_resolved(tmp_path: Path, mocker: MockFixture, capsys: pytest.CaptureFixture[str]) -> None:
```
kemzeb, line 15: We actually have a `test_pipdeptree.py` file that we can throw this test in so let's add it there (the name of this file is a relic from older versions where all the code existed in a single module called `pipdeptree`). Let's add it right under the test `test_main()`.

tests/test_main.py
```diff
@@ -0,0 +1,38 @@
...
+    cmd = ["", "--python", "auto"]
+    mocker.patch("pipdeptree.__main__.sys.argv", cmd)
+
+    mocker.patch("pipdeptree._detect_env.detect_active_interpreter", return_value=str(tmp_path))
+    mocker.patch("pipdeptree._detect_env.os.environ", {"VIRTUAL_ENV": str(tmp_path)})
+    mocker.patch("pipdeptree._detect_env.Path.exists", return_value=True)
```
kemzeb, line 21: Since we are now mocking `detect_active_interpreter()` it looks like we can throw out these two patch calls.

tests/test_main.py
```diff
@@ -0,0 +1,38 @@
...
+        stderr="",
+    )
+
+    main()
+
+    captured = capsys.readouterr()
+
+    combined_output = captured.out + captured.err
+
+    assert f"(resolved python: {tmp_path!s}" in combined_output
```
kemzeb, line 38: Instead of having to perform string concatenation here, I think it would be better if we do something like this instead:
```suggestion
    captured = capsys.readouterr()
    assert not captured.out
    assert captured.err.startswith(f"(resolved python: {tmp_path!s}")
```

kemzeb (changes requested):

tests/test_pipdeptree.py
```diff
@@ -15,3 +18,23 @@ def test_main() -> None:
...
+    mock_subprocess_run.return_value = CompletedProcess(
+        args=["python", "-c", "import sys; print(sys.path)"],
+        returncode=0,
+        stdout=valid_sys_path,
+        stderr="",
+    )
+
+    result = script_runner.run(["pipdeptree", "--python", "auto"])
+    assert result.stderr
+    assert result.stderr.startswith(f"(resolved python: {tmp_path!s}")
```
kemzeb, line 40: Let's change this test to the following and I think we should be good to go (you will need to add a couple more imports):
```suggestion
def test_main_log_resolved(tmp_path: Path, mocker: MockFixture, capsys: pytest.CaptureFixture[str]) -> None:
    mocker.patch("sys.argv", ["", "--python", "auto"])
    mocker.patch("pipdeptree.__main__.detect_active_interpreter", return_value=str(tmp_path))
    mock_subprocess_run = mocker.patch("subprocess.run")
    valid_sys_path = str([str(tmp_path)])
    mock_subprocess_run.return_value = CompletedProcess(
        args=["python", "-c", "import sys; print(sys.path)"],
        returncode=0,
        stdout=valid_sys_path,
        stderr="",
    )

    main()
    
    captured = capsys.readouterr()
    assert captured.err.startswith(f"(resolved python: {tmp_path!s}")
```

kemzeb (changes requested):

kemzeb (approved):

PR #429, opened 2024-11-24
Implement --path argument

### LINKED ISSUE #408: support non-default /site-packages/
### Describe the feature

Like `pip freeze --path=<...>` allows so that packages installed from an old version of Python are supported.

kemzeb:
Resolves #408.

### DIFF
--- a/src/pipdeptree/_cli.py
+++ b/src/pipdeptree/_cli.py
@@ -63,7 +63,7 @@ def build_parser() -> ArgumentParser:
     )
     select.add_argument(
         "--path",
-        help="comma separated list of paths used to restrict where packages should be looked for",
+        help="Passes a path used to restrict where packages should be looked for (can be used multiple times)",
         action="append",
     )
     select.add_argument(
--- a/src/pipdeptree/_cli.py
+++ b/src/pipdeptree/_cli.py
@@ -13,6 +13,7 @@ from .version import __version__
 class Options(Namespace):
     freeze: bool
     python: str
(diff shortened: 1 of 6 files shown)


src/pipdeptree/_cli.py
```diff
@@ -60,6 +61,11 @@ def build_parser() -> ArgumentParser:
             " it can't."
         ),
     )
+    select.add_argument(
+        "--path",
+        help="comma separated list of paths used to restrict where packages should be looked for",
+        action="append",
```
kemzeb, line 67: When choosing either to have a comma-separated list of paths or having users repeat `--path`, I went for the latter.

Even though our other options follow the comma-separated list approach to multiple values, I felt that having something like:

```sh
pipdeptree --paths /home/anon/site-dir-1,/home/anon/site-dir-2
```

Seems less readable when compared to:

```sh
pipdeptree --path /home/anon/site-dir-1 --path /home/anon/site-dir-2
```

... then again maybe it makes sense to stick with comma-separated lists because for consistency? Open to either one.

kemzeb:
I followed the behavior of pip [freeze](https://pip.pypa.io/en/stable/cli/pip_freeze/) when it using to its own `--path` and specifically observed how it behaves with `--local` and `--user` as well as with virtual envs.

gaborbernat (dismissed):

kemzeb:
Whoops, need to update some documentation

gaborbernat (approved):

PR #443, opened 2025-01-26
Migrate tox config from ini to toml

kemzeb:
This also replaces our tox-ini-fmt pre-commit hook with tox-toml-fmt.


.pre-commit-config.yaml
```diff
@@ -14,11 +14,6 @@ repos:
     hooks:
       - id: codespell
         additional_dependencies: ["tomli>=2.0.1"]
-  - repo: https://github.com/tox-dev/tox-ini-fmt
```
gaborbernat, line 17: Replace with tox-toml-fmt

gaborbernat (changes requested):

gaborbernat (approved):

PR #444, opened 2025-01-26
Drop support for python 3.8

kemzeb:
Drops support for python 3.8 since it is EOL. The code refactoring is due to ruff's formatting rules for 3.9.


pyproject.toml
```diff
@@ -68,7 +67,6 @@ build.hooks.vcs.version-file = "src/pipdeptree/version.py"
 version.source = "vcs"
 
 [tool.ruff]
-target-version = "py38"
```
kemzeb, line 71: According to ruff's [docs](https://docs.astral.sh/ruff/settings/#target-version), they recommend using  `project.requires-python` instead of using this. Since we already have this key, I think it's fine to remove this.

gaborbernat (approved):

kemzeb:
Removed the 3.8 status check, will make a minor bump for this

PR #458, opened 2025-03-04
Update CLI usage for --all by moving it to the "render" group

kemzeb:
It makes more sense in the "render" group as this option moreso affects how the output is rendered. This commit also does the following:
- Update README.md with our current CLI usage
- Update description of --depth to be more accurate

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ In earlier versions, `--json`, `--json-tree` and `--graph-output` options overri
 
 ```bash
 % pipdeptree --help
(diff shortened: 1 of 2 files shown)


kemzeb:
Test failures have to do with pypy tests and the required packages it installs... looks like `readline` is no longer included in pypy-3.10?

kemzeb:
I think the culprit is in this recent changelog: https://doc.pypy.org/en/latest/release-v7.3.19.html#python-3-10

> Remove distinfo around vendored readline module, to be more compatible with CPython.

`readline` is no longer accessible via metadata in pypy-3.10.19

gaborbernat (approved):
Assuming you fix the CI

gaborbernat (approved):

PR #460, opened 2025-03-10
Remove unused graphviz library code

kemzeb:
We currently expect graphviz>=0.20.3, so the try-except branch that was removed in this snapshot shouldn't ever execute.


kemzeb:
Need to update the README doc during my free time since it mentions the deprecation warning.

gaborbernat (approved):

PR #465, opened 2025-03-20
Support for License-Expression field

PalmtopTiger:
Allows to display a license for projects with a new License-Expression field. For example, pipdeptree itself.

https://peps.python.org/pep-0639/#add-license-expression-field

### DIFF
--- a/src/pipdeptree/_models/package.py
+++ b/src/pipdeptree/_models/package.py
@@ -40,12 +40,14 @@ class Package(ABC):
             return self.UNKNOWN_LICENSE_STR
 
         license_strs: list[str] = []
-        classifiers = dist_metadata.get_all("Classifier", [])
+        if license_str := dist_metadata.get("License-Expression"):
+            license_strs.append(license_str)
 
+        classifiers = dist_metadata.get_all("Classifier", [])
         for classifier in classifiers:
             line = str(classifier)
             if line.startswith("License"):
-                license_str = line.split(":: ")[-1]
+                license_str = line.rsplit(":: ", 1)[-1]
                 license_strs.append(license_str)
 
         if len(license_strs) == 0:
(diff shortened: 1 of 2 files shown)


gaborbernat (changes requested):
You'll need to add test for your change 👍

PalmtopTiger:
Done

kemzeb (changes requested):
Thanks for your contribution! The implementation and tests appear fine, but from testing I did notice the following:
```
pipdeptree --license -d 0
. . .
tox==4.24.1 (MIT, MIT License)
. . .
```

We have both the `License-Expression` and a `Classifier` field that represent the same license. I'm not sure if this is fine.

I think we could just use the `License-Expression` field if there are license classifiers also defined. Otherwise use the license classifiers.

Reading [this](https://peps.python.org/pep-0639/#deprecate-license-classifiers) section in the PEP, it explains that "if the `License-Expression` field is present, build tools **MAY** raise an error if one or more license classifiers is included in a `Classifier` field". Not too big of a fan of the use of "MAY", but in my opinion it is the responsibility of project maintainers to define their licenses correctly in their project metadata. To me it would be weird if some licenses were defined in one place and some in another.

PalmtopTiger:
No, it's not fine, judging by the fact that setuptools v77 [doesn't allow](https://github.com/pypa/setuptools/blob/7c859e017368360ba66c8cc591279d8964c031bc/setuptools/dist.py#L430) you to install a package where license expression and license classifier are specified at the same time.

I also think that only license expression should be returned if it is available. That's why I placed the code for getting it above the old code. Since you think the same, I've reworked the code.

kemzeb (approved):
LGTM!

gaborbernat (approved):

PR #476, opened 2025-04-29
Replace `pytest-console-scripts` plugin with smaller in-house solution

### LINKED ISSUE #474: CI failures for 3.9

latest

No

### Describe the problem

See https://github.com/tox-dev/pipdeptree/actions/runs/14704056780/job/41276933886. @kemzeb can you take a look? Thanks!

kemzeb:
Resolves #474.

Due to us using the stdlib `importlib.metadata` and `pytest-console-scripts` using the backport version for Python versions < 3.10, this has caused some problems when `importlib_metadata` upgrades to a version that makes changes which is incompatible with the stdlib version.

After analyzing this plugin and the stdlib version of `importlib.metadata`, I ultimately decided to remove this plugin in favor of using `Distribution.entry_points()` to get the console script.

### DIFF
--- a/tests/test_pipdeptree.py
+++ b/tests/test_pipdeptree.py
@@ -28,7 +28,7 @@ def test_console_script() -> None:
     entry_points = dist.entry_points
     assert len(entry_points) == 1
 
-    if (sys.version_info >= (3, 11)): # pragma: >=3.11
+    if sys.version_info >= (3, 11):  # pragma: >=3.11
         entry_point = entry_points["pipdeptree"]
     else:
         entry_point = entry_points[0]
--- a/tests/test_pipdeptree.py
+++ b/tests/test_pipdeptree.py
@@ -27,7 +27,11 @@ def test_console_script() -> None:
 
     entry_points = dist.entry_points
     assert len(entry_points) == 1
-    entry_point = entry_points[0]
+
+    if (sys.version_info >= (3, 11)): # pragma: >=3.11
+        entry_point = entry_points["pipdeptree"]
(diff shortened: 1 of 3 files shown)


tests/test_pipdeptree.py
```diff
@@ -1,26 +1,41 @@
...
 def test_main() -> None:
     check_call([sys.executable, "-m", "pipdeptree", "--help"])
 
 
-def test_console(script_runner: ScriptRunner) -> None:
-    result = script_runner.run(["pipdeptree", "--help"])
-    assert result.success
+def test_console_script() -> None:
+    try:
+        dist = metadata.distribution("pipdeptree")
```
kemzeb, line 24: I think it's safe to assume that OS packagers will have an environment that contains pipdeptree metadata?

tests/test_pipdeptree.py
```diff
@@ -1,26 +1,45 @@
...
+    except Exception as e:  # noqa: BLE001 # pragma: no cover
+        pytest.fail(f"Unexpected error when retrieving pipdeptree metadata: {e}")
+
+    entry_points = dist.entry_points
```
kemzeb, line 28: Just want to mention, due to the [spec](https://packaging.python.org/en/latest/specifications/recording-installed-packages/#the-entry-points-txt-file), it's possible that the `entry_points.txt` file may not exist. This maybe important to consider when Linux distro packagers run our test suite.

I'm unsure what installers are out there that don't create this file..?

gaborbernat (approved):

PR #488, opened 2025-07-05
Implement --exclude-dependencies

### LINKED ISSUE #453: Implement `--exclude-dependencies` to make `--exclude` exclude a package's dependencies
### Describe the feature

Related discussions are [here](https://github.com/tox-dev/pipdeptree/discussions/434) and [here](https://github.com/tox-dev/pipdeptree/issues/452).

Currently, `--exclude` does its job of excluding packages from the dependency tree. This is fine, but when there are times when we also want to remove their dependencies from the output the current implementation doesn't make this possible.

This issue proposes an `--exclude-dependencies` flag that will have pipdeptree take this additional step of removing dependencies from the output.

In the future major versions, it may make sense to just make this the default behavior of `--exclude`. But to avoid a breaking change, it should just be a flag for now.

kemzeb:
Resolves #453.

Much harder than I expected to implement, but I think my logic and implementation checks out. The algorithm is the following:

- Take the _exclude_ set from CLI input and resolve it to actual nodes in the graph
- Determine all possible candidates for exclusion by performing DFS
  - So we add the direct and transitive dependencies of the exclude nodes we gathered in Step 1 into a `candidates` set, but **we aren't sure yet** if they should be excluded since a non-excluded node maybe using it. **This is the reason** why we have more steps below
- Create a reversed graph so that we know all the nodes that depend on a specific node
- Iterate over `candidates`, **counting** the number of dependents for each node using the reversed graph
- Now iterate over the explicitly-excluded nodes that we created in Step 1, where for each node:
  - Add it to the **new excluded set**
  - Iterate over their children. For each child, get the dependent count of that child and decrement it by 1
    - If the count equals 0, we haven't added it to the new set, and it isn't apart of the original resolved set created in Step 1, this child is a candidate for exclusion so let's add it (and in effect, also add their applicable children)
    - If this specific branch is never reached, it means that **their is a non-excluded node that has an outgoing edge to child**. We should not remove it
- Return the **new excluded set**, which will contain both the explicitly-excluded nodes and the dependency nodes associated to them


Shread1234:
Sorry to resurrect a merged PR, but I just want to say thank you! I've wanted this for a long time, but never had the drive to open a PR myself. Great work! 😄

kemzeb:
No problem!

PR #519, opened 2025-11-09
Implement --output to centralize choosing an output format

### LINKED ISSUE #349: Streamline graph output (graphviz, mermaid, etc.)
### Describe the feature

Hi, first of all big thanks for creating pipdeptree :)

I was looking into adding support for D2 graph generation, and I think it would be better to avoid adding another output option at this time.
Maybe streamlining the current graphviz and mermaid options under a common option (for example --output mermaid, --output graphviz_dot, etc.) and then supporting new formats is a workable enhancement ?

kemzeb:
Resolves #349.

### DIFF
--- a/src/pipdeptree/__main__.py
+++ b/src/pipdeptree/__main__.py
@@ -66,11 +66,7 @@ def main(args: Sequence[str] | None = None) -> int | None:
                 warning_printer.print_single_line(str(e))
             return _determine_return_code(warning_printer)
 
-    try:
-        render(options, tree)
-    except ValueError as e:
-        print(e, file=sys.stderr)  # noqa: T201
-        return 1
+    render(options, tree)
 
     return _determine_return_code(warning_printer)
 


src/pipdeptree/__main__.py
```diff
@@ -67,11 +66,21 @@ def main(args: Sequence[str] | None = None) -> int | None:
...
+        render(options, tree)
+    except ValueError as e:
+        print(e, file=sys.stderr)  # noqa: T201
+        return 1
 
     return _determine_return_code(warning_printer)
 
 
+def _is_text_output(options: Options) -> bool:
+    if any([options.json, options.json_tree, options.graphviz_format, options.mermaid]):
```
kemzeb: I added mermaid here so we won't print warnings for it. Is this fine?

To be honest I'm not sure why we have been hide warnings for these formats as we write them to stderr anyways

gaborbernat (dismissed):

gaborbernat (approved):

Issue #335, opened 2024-03-20
Depend on pip


pawamoy:
pipdeptree depends on pip:

```python
Traceback (most recent call last):
  File "/media/data/dev/markdown-exec/.venv/bin/pipdeptree", line 5, in <module>
    from pipdeptree.__main__ import main
  File "/media/data/dev/markdown-exec/.venv/lib/python3.11/site-packages/pipdeptree/__main__.py", line 10, in <module>
    from pipdeptree._models import PackageDAG
  File "/media/data/dev/markdown-exec/.venv/lib/python3.11/site-packages/pipdeptree/_models/__init__.py", line 3, in <module>
    from .dag import PackageDAG, ReversedPackageDAG
  File "/media/data/dev/markdown-exec/.venv/lib/python3.11/site-packages/pipdeptree/_models/dag.py", line 12, in <module>
    from .package import DistPackage, ReqPackage, pep503_normalize
  File "/media/data/dev/markdown-exec/.venv/lib/python3.11/site-packages/pipdeptree/_models/package.py", line 10, in <module>
    from pip._vendor.pkg_resources import Requirement  # noqa: PLC2701
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'pip'
```

...so I would suggest to add `pip` as a dependency, in pyproject.toml. It is less and less common to have pip automatically installed in virtualenvs, so this shouldn't be expected anymore that pip is always installed alongside a package :slightly_smiling_face: Happy to send a PR!

Related: https://github.com/tox-dev/pipdeptree/issues/331

gaborbernat:
cc @kemzeb

kemzeb:
Seems reasonable enough; feel free to open a PR!

austinpray-mixpanel:
Opened a PR to lax up the pip version range to support use case of using pipdeptree to revive projects stuck on ancient versions of pip 👍 https://github.com/tox-dev/pipdeptree/pull/340

Issue #150, opened 2021-07-27
pipdeptree --freeze fails with AttributeError exception (no attribute 'editable')


phries:
Hello! I came across this error in my typical usage pattern, where `--freeze` is now failing. Maybe the internal pip API changed?

Console output:
```
$ python3 -m pipdeptree --freeze
Warning!!! Possibly conflicting dependencies found:
* launchpadlib==1.10.6
 - testresources [required: Any, installed: ?]
------------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.6/dist-packages/pipdeptree.py", line 878, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.6/dist-packages/pipdeptree.py", line 872, in main
    render_text(tree, args.all, args.freeze)
  File "/usr/local/lib/python3.6/dist-packages/pipdeptree.py", line 510, in render_text
    lines = flatten([aux(p) for p in nodes])
  File "/usr/local/lib/python3.6/dist-packages/pipdeptree.py", line 510, in <listcomp>
    lines = flatten([aux(p) for p in nodes])
  File "/usr/local/lib/python3.6/dist-packages/pipdeptree.py", line 498, in aux
    node_str = node.render(parent, frozen)
  File "/usr/local/lib/python3.6/dist-packages/pipdeptree.py", line 107, in render
    return self.render_as_root(frozen)
  File "/usr/local/lib/python3.6/dist-packages/pipdeptree.py", line 141, in render_as_root
    return self.__class__.frozen_repr(self._obj)
  File "/usr/local/lib/python3.6/dist-packages/pipdeptree.py", line 113, in frozen_repr
    fr = frozen_req_from_dist(obj)
  File "/usr/local/lib/python3.6/dist-packages/pipdeptree.py", line 81, in frozen_req_from_dist
    return FrozenRequirement.from_dist(dist)
  File "/usr/local/lib/python3.6/dist-packages/pip/_internal/operations/freeze.py", line 257, in from_dist
    req, editable, comments = _get_editable_info(dist)
  File "/usr/local/lib/python3.6/dist-packages/pip/_internal/operations/freeze.py", line 165, in _get_editable_info
    if not dist.editable:
[...]

PR #552, opened 2026-03-10
♻️ refactor(deps): drop pip dependency, use stdlib for metadata

gaborbernat:
pipdeptree depended on `pip._internal` APIs that are explicitly not part of pip's public interface. Every pip release risked breakage — pip 21.1.1 already required a compatibility fix (#150), and the project had to isolate all pip usage into a single module just to contain the blast radius (#402). 🔧 The dependency also blocked pipdeptree from working in pip-less environments (#335) and added ~50MB to installation size.

This replaces all `pip._internal` imports with a new `_parser` package built entirely on stdlib (`importlib.metadata`, `json`, `urllib`). It implements PEP 610 `direct_url.json` parsing with full validation, editable install detection for both modern PEP 660 and legacy `.egg-link` files, and VCS requirement string generation for git, hg, svn, and bzr. ✨ The VCS module is split into per-backend files with shared utilities, matching pip's output format for freeze-style requirement strings including credential redaction, SCP-style git URL normalization, and subdirectory fragments.

The `pip>=25.2` dependency is removed from `pyproject.toml` and `_freeze.py` is deleted entirely. pipdeptree now works independently of pip's release cycle and in any Python environment with `importlib.metadata` available. All existing output formats are preserved.


src/pipdeptree/_parser/_vcs/_git.py
```diff
@@ -0,0 +1,126 @@
...
+    """Find git repository root from any subdirectory."""
+    try:
+        repo_root = subprocess.run(
+            ["git", "rev-parse", "--show-toplevel"],  # noqa: S607
+            cwd=location,
+            capture_output=True,
+            text=True,
+            check=True,
+            timeout=5,
+        ).stdout.strip()
```
kemzeb: The `GIT_DIR` and `GIT_WORK_TREE` environment variables, if set, would not honor the `location`.

Found this while reviewing [pip](https://github.com/pypa/pip/blob/b92e6d1ecba110f6f959d194db0bc045da60c340/src/pip/_internal/vcs/git.py#L73-L75)'s implementation and the issue it links there, where it unsets these environment variables. This can be resolved by:
- Performing an `os.environ.copy()`
- Popping the unwanted env vars from the copy
- Pass them to the subprocess

This will need to be done for every subprocess call

src/pipdeptree/_parser/_vcs/_bzr.py
kemzeb: nit: Since `_parser/` is considered private, does `_vcs` and its modules need to also have a leading underscore?

src/pipdeptree/_parser/_format.py
```diff
@@ -0,0 +1,114 @@
...
+    direct_url = get_direct_url(distribution)
+    if direct_url:
+        if direct_url.is_editable():
+            location = url_to_path(direct_url.url)
+            return _format_editable(location, distribution.metadata["Name"], distribution.version)
+        return format_requirement(direct_url, distribution.metadata["Name"])
+    if egg_link := find_egg_link(distribution.metadata["Name"]):
+        location = read_egg_link_location(egg_link)
+        return _format_editable(location, distribution.metadata["Name"], distribution.version)
+    name = distribution.metadata["Name"]
```
kemzeb: nit: Define the `name` variable earlier:

```suggestion
    direct_url = get_direct_url(distribution)
    name = distribution.metadata["Name"]
    if direct_url:
        if direct_url.is_editable():
            location = url_to_path(direct_url.url)
            return _format_editable(location, name, distribution.version)
        return format_requirement(direct_url, name)
    if egg_link := find_egg_link(name):
        location = read_egg_link_location(egg_link)
        return _format_editable(location, name, distribution.version)
```

kemzeb (changes requested):
Some feedback, still need to dive into the tests as well.

Also looking to pass through the pip implementation and PEP specs a couple more times to try and catch anything

kemzeb (approved):
LGTM

PR #572, opened 2026-04-16
Add SECURITY.md

jrwe:
Adds SECURITY.md with vulnerability reporting


kemzeb:
We already have a security policy here: https://github.com/tox-dev/pipdeptree/blob/main/.github/SECURITY.md

PR #580, opened 2026-05-05
⚡ perf(extras): make --extras tractable on real environments

### LINKED ISSUE #559: Make `--extras`/`-x` the default
### Describe the feature

The only reason it’s not is for backwards compatibility, but it makes the tool strictly more useful.

Maybe we can warn people who specify neither `--extras` or `--no-extras` that `--extras` will be the default and they’ll need to specify `--no-extras` to get the old behavior in the future.

gaborbernat:
Issue #559 reported that `--extras` turned a one-second invocation into an eighteen-second one and inflated the rendered tree from 700 to 20,000 lines on a real codebase. The cause was three compounding sources in the satisfaction check: PEP 508 parsing of every `Requires-Dist` on every call, no memoization across the top-level `(pkg, extra)` queries iterated by `_collect_satisfied_extras`, and unbounded edge appends across resolution rounds. A separate hotspot in the renderer reread each package's METADATA file once per emitted edge, which dominated the wall-clock cost on large environments. The active-iff-installed semantic itself is correct because it mirrors the runtime `try: import optional_dep` idiom Python libraries use; the goal was to keep that meaning while making the implementation match it in cost.

## How resolution works now

Every `(pkg, extra)` query consults a three-tier cache before doing any work, and every result lands in the cache appropriate to its provenance:

```mermaid
flowchart TD
    Q["is_satisfied(pkg, extra)"]:::start --> L{lookup}
    L -->|hit global cache| R1[return cached]:::cached
    L -->|hit scope cache| R2["return cached, used=True"]:::tentative
    L -->|in_progress| R3["return True, used=True (cycle)"]:::tentative
    L -->|miss| W[push frame on stack]:::work
    W --> S[advance one step]:::work
    S -->|sub-extra to resolve| L
    S -->|all subs done| F{used cycle?}
    F -->|yes| SC[scope_cache - per query]:::tentative
    F -->|no| GC[global cache - across queries]:::cached
    SC --> POP[pop frame]:::work
    GC --> POP
    classDef start fill:#2980b9,color:#fff,stroke:#1c5980
    classDef cached fill:#27ae60,color:#fff,stroke:#1e8449
    classDef tentative fill:#8e44ad,color:#fff,stroke:#5b2c6f
    classDef work fill:#e67e22,color:#fff,stroke:#a04000
```

⚡ Splitting the cache this way preserves the original "fresh `visited` per top-level call" semantics, where a node's truth value can legitimately differ depending on the cycle entry point, while still amortizing the acyclic majority of work across the whole environment. The walk itself is iterative on an explicit frame stack:

```mermaid
flowchart LR
    subgraph Before
        A[recursive _resolve]:::work -->|"chain > 1000"| AE[RecursionError]:::fail
        A -->|large SCC| AE
    end
    subgraph After
        B[stack-based _resolve]:::work -->|any depth| BE[resolves]:::cached
    end
    classDef work fill:#e67e22,color:#fff,stroke:#a04000
    classDef fail fill:#e74c3c,color:#fff,stroke:#922b21
    classDef cached fill:#27ae60,color:#fff,stroke:#1e8449
```

`DistPackage` parses each `Requires-Dist` exactly once into a shared cached property feeding both the mandatory iterator and the extras index, with canonical dep keys precomputed so the hot path skips `canonicalize_name`. Edge deduplication in `_resolve_extras` keeps malformed metadata with repeated `Requires-Dist` entries from producing repeated edges. `DistPackage.version` is cached too, so the renderer no longer reparses METADATA for every emitted node.

## Real-world impact

A virtualenv assembled from the packages sparrowt mentioned in the issue (Django, asgiref, pytest, pytest-asyncio, Sphinx, pylint-django, plus the broader Django/REST/Celery/Sphinx/scientific Python ecosystems they would coexist with):

| Environment | Operation | Output | Baseline | This PR | Speedup |
|---|---|---:|---:|---:|---:|
| 268 pkgs (Django + ML + REST stack) | `--extras` | 79 433 lines | 7.71 s | **0.44 s** | **17.5x** |
| 268 pkgs (same env) | `--no-extras` | 2 523 lines | 0.49 s | **0.25 s** | **2.0x** |
| 174 pkgs (Django + Sphinx + tooling) | `--extras` | 6 968 lines | 1.07 s | **0.91 s** | 1.2x |

The 268-package environment matches sparrowt's reported scale (their `--extras` produced ~20 000 lines and ran in ~18 seconds; this env's render is ~4x larger and now runs in 0.44 seconds).

## Synthetic stress tests

Fully connected `(pkg, extra)` SCC where every extra is satisfied — the worst case the active-iff-installed semantic can produce:

| Workload (libs / extras / deps) | Baseline | This PR | Speedup |
|---|---:|---:|---:|
| 50 / 3 / 2 | 2562 ms | 41 ms | **62x** |
| 100 / 4 / 3 | 38 032 ms | 370 ms | **103x** |
| 200 / 5 / 4 | `RecursionError` | 3108 ms | resolves |

Acyclic deep chains (the asgiref → pytest → Sphinx pattern from the issue):

| Library count | Baseline | This PR | Speedup |
|---|---:|---:|---:|
| 1000 | 278 ms | 74 ms | **3.8x** |
| 5000 | (untested) | 414 ms | scales |
| 1500-deep chain | `RecursionError` | passes | resolves |

📝 Documentation across the four Diataxis sections gained an explanation of the active-iff-importable semantic, an annotated how-to example, an enriched CLI help string, and a tutorial pointer; that rule used to surface only in the source. No CLI semantics change, so existing invocations behave identically.

Closes #559.

### DIFF
--- a/docs/explanation.rst
+++ b/docs/explanation.rst
@@ -65,9 +65,47 @@ Circular dependency detection uses cycle detection on the directed dependency gr
         style X fill:#e67e22,color:#fff
         style Y fill:#e67e22,color:#fff
 
+Optional dependencies (extras)
+------------------------------
+
+When you pass ``--extras``, pipdeptree augments the tree with optional dependency edges. Two kinds
+of extras are surfaced:
+
+1. **Explicitly requested extras.** When a parent's metadata records a dependency like
+   ``oauthlib[signedtoken]``, the deps gated behind oauthlib's ``signedtoken`` extra get
+   added under oauthlib in the tree, annotated with ``extra: signedtoken``.
+
+2. **Active extras.** Python packaging metadata never records *why* a package was installed,
+   only what it could require. pipdeptree therefore considers an extra "active" when every dep
+   that extra would have requested is already installed -- the same heuristic Python libraries use
+   at runtime to decide whether a feature is available (``try: import optional_dep``). When an
+   extra is active, pipdeptree adds the same edges it would have added if the extra had been
+   requested explicitly.
+
+Both kinds propagate. If activating ``A[x]`` pulls in ``B[y]``, pipdeptree also adds the deps
+``B[y]`` would gate. Because metadata cannot tell us that ``B`` was installed for some other
+reason, the same package may appear under multiple parents through this mechanism.
+
+.. mermaid::
+
+    flowchart TD
+        A[App] -- requires --> O[oauthlib]
+        A -- "requires<br/>oauthlib[signedtoken]" --> O
+        O -. "extra: signedtoken" .-> C[cryptography]
+        O -. "extra: signedtoken" .-> P[pyjwt]
+        style C fill:#8e44ad,color:#fff
+        style P fill:#8e44ad,color:#fff
+
+To leave optional edges out entirely, omit ``--extras`` (the default). Edges added through
+extras are always annotated with the originating extra, so they remain distinguishable from
+mandatory dependencies in every output format.
+
 Limitations
 -----------
 
 - pipdeptree only sees packages that are already installed. It cannot predict what a ``pip install`` will do.
 - If you need a dependency resolver that works without installing packages first, consider :pypi:`uv`.
 - Extra/optional dependencies are not shown by default; use ``--extras`` to include them.
+- ``--extras`` cannot reconstruct extras that were requested only on the command line
+  (e.g. ``pip install foo[dev]`` where ``foo`` is itself top-level), because that information
+  is never persisted into installed package metadata.
--- a/docs/how-to/usage.rst
+++ b/docs/how-to/usage.rst
@@ -320,3 +320,16 @@ Show optional (extras) dependencies in the tree with ``--extras`` (``-x``):
 
 Without ``--extras``, only mandatory dependencies are shown. Packages that declare optional dependency groups (extras)
 will have those additional dependencies included when this flag is set.
+
+Edges added through an extra are annotated with that extra's name:
+
+.. code-block:: console
+
+    $ pipdeptree --extras --packages oauthlib
+    oauthlib==3.0.0
+    ├── cryptography [required: Any, installed: 2.7, extra: signedtoken]
+    └── pyjwt [required: >=1.0.0, installed: 1.7.1, extra: signedtoken]
+
+An extra is included not only when a parent explicitly requested it (e.g. ``oauthlib[signedtoken]``)
+but also when every dependency that the extra would require is already installed in the
+environment. See :doc:`/explanation` for the rationale.
--- a/docs/tutorial/getting-started.rst
+++ b/docs/tutorial/getting-started.rst
@@ -142,5 +142,7 @@ Limit the tree depth:
 Next steps
 ----------
 
-See :doc:`/how-to/usage` for filtering, virtualenv support, and warning control. See :doc:`/how-to/output-formats` for
-all available output formats including JSON, Mermaid, and Graphviz.
+See :doc:`/how-to/usage` for filtering, virtualenv support, warning control, and how to surface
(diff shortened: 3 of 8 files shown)


kemzeb:
I'll review by the end of this week

kemzeb:
Looking good so far, plan to finish reviewing and testing tomorrow

kemzeb:
Apologies it's been quite a busy work week, I will review on Friday. It looks good, just need to dive deeper into `_ExtrasResolver`

kemzeb (commented):
I've been digging into `_ExtrasResolver`; I'm looking to understand the difference between global cache and the scope cache? I added a `print()` to the `lookup()` and ran it against my environment that has nested extras in its tree and noticed that the scope cache is always empty. Observing the code further, it looks like:
* The only time the scope cache is written to is if it's requirement key-extra name pair is considered "used"
  * It's only ever considered used when `frame.used_assumption = True` (this is done in `fold_pending()`
* `_lookup()` appears to be the only function call in `_advance()` that can change it to "used"
  * Observing the `_lookup()` call, it will only be considered used if either:
    * The requirement key-extra name pair exists in the scope cache
    * The "in-progress" cache contains the requirement key-extra name pair (also, when adding `print(len(self._in_progress))` to `_lookup()`, I see it have a size of either 0 or 1)

So from my observations it appears that the scope cache won't be built during runtime. Would it make since to just utilize the global cache? Forgive me if I have missed something in the design

gaborbernat:
While reviewing `_ExtrasResolver`, I investigated whether the per-query `scope_cache` is exercised by any real-world environment. The scope cache quarantines cycle-tainted results (those that relied on the optimistic `True` assumption when hitting an `_in_progress` node) to prevent a global cache from serving wrong answers when the same node is queried from a different entry point.

### The cycle scenario it guards against

For the scope cache to activate, the extras graph needs a cycle: `A[x]` requires `B[y]`, and `B[y]` (transitively) requires `A[x]`. This means two packages each declare the other as an extras-gated optional dependency in their `Requires-Dist` metadata. Here's a concrete example where global caching of cycle-tainted results would produce a wrong answer:

```
A[x] requires B[y]
B[y] requires C[z], missing-pkg  (missing-pkg NOT installed)
C[z] requires B[y]              (cycle)
```

When resolving `A[x]`, `C[z]` resolves to `True` (because `B[y]` is optimistically assumed `True` via the cycle). If that `True` were cached globally, a later top-level query for `C[z]` would return `True` — but the correct answer is `False` (since `B[y]` fails due to `missing-pkg`).

### No real-world package has cyclic extras

Checked 30+ packages across every major Python ecosystem (Airflow, Dask, IPython, FastAPI, Celery, Google Cloud, Prefect, Setuptools, Django, Sphinx, Pandas, Sentry SDK, scikit-learn, Jupyter, OpenTelemetry, and more) via PyPI metadata — none have cyclic extras dependencies. Extras deps flow from higher-level to lower-level packages; no real package says "my optional feature Y needs your optional feature X" when the other package already says the reverse.

### Simplification

Replaced the scope cache with a simpler approach: **discard cycle-tainted results** (don't cache them at all). If a cycle is hit, the path is re-walked on the next top-level query. Since no real environment has extras cycles, this has zero performance cost in practice.

kemzeb (approved):
Don't see further glaring issues. Will cut a release

kemzeb:
I will keep the linked issue open as it's moreso a feature request to make rendering extras the default. There is also conversations about showing explicit extras only.

Looking at our current logic this may be doable (i.e. introduce a flag that has us not call `_collect_satisfied_extras()`). Maybe it we cut a breaking release, it may make more sense to have explicit extras by default and introduce a `--extras-collect-satisfied` or similar name for users who are looking for this additional info.

PR #598, opened 2026-06-08
ci: Test against pypy-3.11

kemzeb:
I think we'll start testing against the latest python version that pypy can support.


gaborbernat (approved):
LGTM

PR #606, opened 2026-06-30
Validate archive hash parts

CodingFeng101:
## What changed

This validates both sides of legacy `archive_info.hash` values from `direct_url.json`.

Values such as `=abc123` and `sha256=` are now rejected as invalid instead of producing an `ArchiveInfo.hashes` entry with an empty algorithm or digest.

## Why

PEP 610's legacy `archive_info.hash` is a `name=value` string. Accepting an empty hash name or empty digest can leak malformed install metadata into `pipdeptree`'s direct URL formatting path instead of treating it as invalid metadata.

## Validation

- `python -m pip install -e .`
- `python -m pytest tests\_parser\test_direct_url.py -q` -> 54 passed
- `python -m ruff check src\pipdeptree\_parser\direct_url.py tests\_parser\test_direct_url.py`
- `python -m ruff format --check src\pipdeptree\_parser\direct_url.py tests\_parser\test_direct_url.py`
- `git diff --check`

I also tried `python -m pytest tests\_parser -q`; it gets past the direct_url tests but this Windows environment is missing the `fp` fixture used by the VCS parser tests and has existing POSIX path expectations in editable-format tests, so I kept the validation scoped to the parser module touched here.

### DIFF
--- a/src/pipdeptree/_parser/direct_url.py
+++ b/src/pipdeptree/_parser/direct_url.py
@@ -119,7 +119,8 @@ def _parse_archive_info(archive_data: dict) -> ArchiveInfo:
         if not isinstance(hash_value, str):
             msg = "archive_info.hash must be a string"
             raise DirectUrlValidationError(msg)
-        if "=" not in hash_value or len(hash_value.split("=", 1)) != 2:
+        algo, _, digest = hash_value.partition("=")
+        if not algo or not digest:
             msg = f"invalid archive_info.hash format: {hash_value!r}"
             raise DirectUrlValidationError(msg)
     hashes: dict[str, str] = {}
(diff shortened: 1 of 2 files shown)


kemzeb (approved):
Thanks for your contribution, [reading the spec](https://packaging.python.org/en/latest/specifications/direct-url-data-structure/#archive-urls) it seems that it does require the `hash` value to be in the correct format.

I setup a test package with metadata that contains an invalid `hash` value to see how pip handles it (using pipdeptree==2.16.0, when we used pip as a dependency), and it appears that they do the same as we do and actually ignore validating the hash value.

I will approve this as I would think if an invalid hash is passed, pip and similar tools that can read the freeze format may fail to parse or install that package. `hash` is a deprecated field anyways so I would suspect modern Python build tools are using `hashes`. We could print out warnings so users can see these validation failures but that's out-of-scope of this PR. Will leave this PR open for a bit in case @gaborbernat has any objections

PR #618, opened 2026-07-13
✨ feat(rust): replace the Python engine with a Rust extension

gaborbernat:
One Rust extension discovers distributions, parses metadata, constructs the dependency graph, and renders the result. Those stages account for most CLI runtime in pipdeptree. The public Python API and CLI keep their supported behavior.

The wheel keeps a six-file Python shim for byte output, notebook display hooks, `python -m pipdeptree`, version attributes, and typing. The replaced Python engine and its fallback switch are gone.

The native boundary follows the [Cargo](https://doc.rust-lang.org/cargo/) and [PyO3](https://pyo3.rs) layout used by [toml-fmt](https://github.com/gaborbernat/toml-fmt). [Meson](https://mesonbuild.com/) owns the [PEP 517](https://peps.python.org/pep-0517/) build through [meson-python](https://mesonbuild.com/meson-python/) and delegates the extension target to Cargo, leaving one Cargo graph for Rust compilation and one Meson graph for wheel assembly.

## Design

```mermaid
flowchart TD
    SHIM["Python shim<br/>CLI · API · notebook · python -m"] --> OPT["options<br/>parse & validate · typed Format"]
    OPT --> ENV["environment<br/>resolve interpreter, paths, PEP 508 markers"]
    ENV --> SRC{"package source"}
    SRC -->|installed| MET["metadata<br/>discover dist-info · parse core metadata<br/>direct URLs · records · licenses · sizes"]
    SRC -->|from-lock| LOCK["lock<br/>PEP 751 pylock.toml"]
    SRC -->|from-index| IDX["index<br/>requirements · pyproject · nab resolve"]
    MET --> GRAPH
    LOCK --> GRAPH
    IDX --> GRAPH
    GRAPH["graph<br/>build · extras · filter · version"] --> RENDER["render<br/>dispatch by Format"]
    RENDER --> OUT["text · rich · freeze · json · json-tree<br/>mermaid · graphviz · summary"]
    ENV -.->|query| PROC["process<br/>subprocess boundary"]
    IDX -.->|nab| PROC
    RENDER -.->|dot| PROC
    classDef boundary fill:#cfe4ff,stroke:#1f6feb,color:#0b1f3a;
    classDef stage fill:#c8e6d0,stroke:#1a7f37,color:#0b2a17;
    classDef source fill:#ffe3a3,stroke:#d29200,color:#3a2c00;
    class SHIM,PROC boundary;
    class OPT,ENV,GRAPH,RENDER,OUT stage;
    class SRC,MET,LOCK,IDX source;
```

One Rust module owns each stage. `environment` resolves the target interpreter and its markers; `metadata`, `lock`, and `index` are the three package sources behind a shared result shape; `graph` builds and filters the tree and resolves versions for uninstalled dependencies; `render` dispatches by the typed output format to a per-format renderer; and `process` is the single subprocess boundary for interpreter discovery, nab resolution, and the [Graphviz](https://graphviz.org/) `dot` pipe. The version-control backends sit behind one trait, so a new backend adds a file rather than a branch, and the render vocabulary lives in the extension, so the Python shim never restates the flag names.

[`pep508_rs`](https://crates.io/crates/pep508_rs) supplies [PEP 440](https://peps.python.org/pep-0440/) and [PEP 508](https://peps.python.org/pep-0508/) semantics. [`serde`](https://serde.rs/) with [`serde_json`](https://crates.io/crates/serde_json) replaces Python serialization; [`toml`](https://crates.io/crates/toml) replaces the Python packaging and TOML dependencies. Rust handles requirement files and PEP 508 input; local and VCS sources; constraints, installed metadata, and [PEP 751](https://peps.python.org/pep-0751/) locks. Discovery keys its lookups with [`rustc-hash`](https://crates.io/crates/rustc-hash) and parses across cores with [`rayon`](https://crates.io/crates/rayon).

The major release makes [`nab-index`](https://pypi.org/project/nab-index/) and [`nab-python`](https://pypi.org/project/nab-python/) base dependencies. PyO3 calls nab for resolution, retaining its resolver and transport boundary without moving packaging or TOML parsing back into Python.

Rust uses [`anstyle`](https://crates.io/crates/anstyle) for rich output. It generates Graphviz source and calls the installed `dot` executable for binary formats. The major release removes the empty `rich`, `graphviz`, and `index` extras.

TTY output defaults to rich text; redirected output plus `TERM=dumb` and `NO_COLOR` retain plain text. Package names use cyan. Versions and satisfied states use green; constraints use bright blue. Metadata and extras use magenta, warnings use yellow, and red applies to errors. Help follows the same terminal detection and color policy.

Discovery and the installed-size sweep run across cores. The engine walks each search path into an ordered candidate list, parses the distributions' metadata in parallel, and folds the results back in order so deduplication stays deterministic; it warms installed sizes in parallel before a size or summary render and keys the graph's name and index maps with a fast non-cryptographic hash. Output stays byte-for-byte identical, and the engine runs 5 to 15 times faster than the former Python engine on the large environment.

## Compatibility

The pre-migration Python suite was the behavior oracle. Each supported behavior now runs through the packaged Python API or CLI, or through the public Rust application boundary. Tests do not import private symbols or widen production visibility. Equivalent inputs use [`pytest.mark.parametrize`](https://docs.pytest.org/en/stable/how-to/parametrize.html) or [`rstest`](https://github.com/la10736/rstest) instead of repeating implementation-shaped cases.

The audit covers metadata discovery and parsing; direct URLs and editables across four VCS backends; graph construction and filters plus extra evaluation; conflicts and cycles, including deep chains; every renderer and exact JSON and text contracts; interpreter and environment selection; CLI validation and warning modes; requirement and pyproject inputs plus index and lock inputs. It found and fixed empty package extras, negative depth parsing, [uv](https://github.com/astral-sh/uv) archive hashes and requested VCS revisions, PyPy environment detection, index precedence, generated TOML omission rules, Graphviz byte output, and active-extra scoping.

The packaged suite passes on CPython 3.10 through 3.14 and PyPy 3.11. Release wheels use [`cp310-abi3`](https://docs.python.org/3/c-api/stable.html) for CPython and native PyPy 3.11 extensions across Linux x86-64, Linux aarch64, macOS arm64, and Windows x86-64.

## Benchmark method

The benchmark uses three generated environments. Each group of 50 packages forms one chain. A package depends on its successor with a Python marker and on the following package through the `speed` extra.

| Size | Packages | Default edges | Edges with active extras | Disk use |
| --- | ---: | ---: | ---: | ---: |
| Small | 100 | 98 | 194 | 1.2 MiB |
| Medium | 1,000 | 980 | 1,940 | 12 MiB |
| Large | 5,000 | 4,900 | 9,700 | 59 MiB |

We ran measurements on an Apple M4 with arm64 macOS 26.5, Python 3.14.6, Rust 1.97.0, and [hyperfine](https://github.com/sharkdp/hyperfine) 1.20.0. We used a release build. Each command had five warmups and 15 measured runs and discarded output through `/dev/null`. Timings include Python startup and extension loading. We report the median followed by inclusive Q1 and Q3 in brackets, all in milliseconds; `Python/Rust` compares medians.

### Small environment

| Workload | Rust ms | Python ms | Python/Rust |
| --- | ---: | ---: | ---: |
| Text, depth 3 | 37.1 [35.2, 38.9] | 83.0 [82.1, 91.2] | 2.24× |
| Rich, depth 3 | 37.0 [35.4, 39.3] | 99.9 [94.5, 117.1] | 2.70× |
| JSON | 36.3 [35.8, 37.5] | 96.1 [92.2, 97.2] | 2.65× |
| Summary JSON | 36.3 [36.0, 37.6] | 102.6 [97.9, 110.2] | 2.83× |
| Active extras JSON | 39.3 [38.5, 39.9] | 98.0 [94.8, 99.1] | 2.49× |
| Metadata, size JSON | 36.4 [36.0, 37.3] | 109.6 [104.4, 116.2] | 3.01× |
| Freeze | 35.9 [35.4, 37.8] | 118.9 [115.5, 123.1] | 3.31× |
| Mermaid | 39.8 [39.4, 40.4] | 88.3 [81.5, 92.7] | 2.22× |
| Graphviz DOT | 36.2 [35.3, 36.6] | 86.8 [84.6, 87.8] | 2.40× |
| Reverse filter, depth 3 | 35.8 [35.4, 36.2] | 88.0 [84.2, 90.4] | 2.46× |

### Medium environment

| Workload | Rust ms | Python ms | Python/Rust |
| --- | ---: | ---: | ---: |
| Text, depth 3 | 46.7 [45.5, 49.4] | 164.8 [160.3, 165.5] | 3.53× |
| Rich, depth 3 | 45.0 [44.7, 45.7] | 178.4 [177.4, 180.0] | 3.96× |
| JSON | 45.5 [44.7, 46.4] | 158.4 [155.4, 160.1] | 3.48× |
| Summary JSON | 53.9 [53.4, 54.7] | 283.1 [279.2, 284.4] | 5.25× |
| Active extras JSON | 45.4 [45.3, 46.5] | 168.2 [164.7, 168.5] | 3.70× |
| Metadata, size JSON | 54.1 [53.8, 54.7] | 291.0 [290.1, 294.8] | 5.38× |
| Freeze | 47.2 [46.4, 48.0] | 334.7 [330.5, 337.4] | 7.09× |
| Mermaid | 45.8 [45.2, 46.5] | 153.7 [151.9, 154.5] | 3.36× |
| Graphviz DOT | 45.7 [45.0, 46.4] | 165.7 [165.1, 167.3] | 3.63× |
| Reverse filter, depth 3 | 44.9 [44.5, 45.5] | 170.8 [165.0, 177.7] | 3.80× |

### Large environment

| Workload | Rust ms | Python ms | Python/Rust |
| --- | ---: | ---: | ---: |
| Text, depth 3 | 88.9 [88.2, 184.5] | 600.8 [576.3, 977.1] | 6.76× |
| Rich, depth 3 | 89.0 [88.3, 91.4] | 556.9 [553.3, 561.8] | 6.26× |
| JSON | 86.1 [86.0, 87.0] | 497.2 [492.2, 511.4] | 5.77× |
| Summary JSON | 127.9 [126.9, 128.5] | 1,128.5 [1,126.3, 1,132.8] | 8.82× |
| Active extras JSON | 88.6 [88.0, 89.4] | 545.6 [543.7, 550.3] | 6.16× |
| Metadata, size JSON | 128.7 [128.2, 129.9] | 1,168.4 [1,163.4, 1,182.8] | 9.08× |
| Freeze | 92.3 [92.0, 92.8] | 1,401.0 [1,393.7, 1,470.8] | 15.18× |
| Mermaid | 86.6 [86.4, 87.3] | 476.5 [474.1, 480.1] | 5.50× |
| Graphviz DOT | 86.5 [86.1, 87.5] | 520.1 [513.6, 524.2] | 6.01× |
| Reverse filter, depth 3 | 84.2 [83.3, 85.1] | 513.4 [510.5, 515.7] | 6.10× |

The engine reads and parses metadata and computes installed sizes across cores, so its advantage grows with the environment. The large matrix ran under background load from a concurrent build, which widens the interquartile ranges on both engines, and in a few cases the median too; the medians are the best observed of two runs and the ratios track the improvement without letting one contended sample decide it.

## Peak memory

We measured peak resident memory in the large environment with three warmups and ten paired fresh processes per workload. We read maximum RSS from `/usr/bin/time -l`. Values are medians in MiB; `Reduction` compares Rust with Python.

| Workload | Rust MiB | Python MiB | Reduction |
| --- | ---: | ---: | ---: |
| Text, depth 3 | 46.4 | 55.8 | 16.9% |
| Rich, depth 3 | 46.3 | 58.4 | 20.8% |
| JSON | 51.0 | 63.6 | 19.8% |
| Summary JSON | 49.8 | 62.2 | 20.0% |
| Active extras JSON | 58.4 | 75.0 | 22.1% |
| Metadata, size JSON | 58.6 | 77.8 | 24.7% |
| Freeze | 47.3 | 58.3 | 18.8% |
| Mermaid | 47.1 | 59.0 | 20.3% |
| Graphviz DOT | 47.1 | 58.1 | 19.0% |
| Reverse filter, depth 3 | 46.3 | 57.8 | 19.8% |

Rust used 16.9% to 24.7% less peak RSS than Python across all ten workloads.

## Profile

We collected 8,383 main-thread samples from a symbolized ten-second macOS run of repeated large Mermaid renders. Discovery accounted for 5,161 samples, including 4,258 in metadata file opens and 679 in a redundant directory check. Rust reuses the file type from each directory entry instead of issuing a separate stat call for every metadata directory. The engine now spreads those metadata reads across cores behind a deterministic merge and warms the installed-size sweep in parallel, so neither bounds a single thread.

Package objects retain metadata requested by the output. Graph construction consumes raw requirements and uses the parsed PEP 508 name instead of two copied strings. Filtered graphs cache expanded child lists and parent counts. Flat JSON writes to its output buffer; rich text renders from graph data; Mermaid and DOT sort vectors and append into one output string.

The first computed JSON path calculated unique dependency sets for `--computed size`, producing quadratic graph work unrelated to size. Rust computes those sets for `unique-deps`. Summary JSON bypasses the text summary pass and uses one backtracking path set instead of cloning the path for every edge.

We ran Python [`cProfile`](https://docs.python.org/3/library/profile.html) over 20 large Mermaid renders: 100,006,605 calls took 32.556 seconds. Distribution discovery took 17.731 seconds, graph construction 14.139 seconds, requirement parsing 9.275 seconds, and metadata parsing and reads about 6.3 seconds each. The native parser removes that Python object and parsing overhead; computed size still requires file metadata calls in both engines.

## Validation

The final packaged Python suite has 90 API and CLI cases, including documentation cases, at 100% coverage. The 372 Rust public-API tests cover 100% of 4,827 executable lines and 592 functions. Nine generated pytest doctests execute the packaged CLI against deterministic metadata and validate documented output. The suite passes on CPython 3.10 through 3.14 and PyPy 3.11. Type checks, strict HTML docs, package metadata and wheel-content checks, both [Clippy](https://github.com/rust-lang/rust-clippy) feature configurations, and the repository hook suite pass.

```console
cargo fmt --all
cargo clippy --all-targets --no-default-features -- -D warnings
cargo clippy --all-targets --all-features -- -D warnings
cargo llvm-cov --no-default-features --test public_api --fail-under-lines 100 --fail-under-functions 100
tox run -e 3.10,3.11,3.12,3.13,3.14,pypy-3.11,type,docs,pkg_meta
prek run --all-files
```


kemzeb:
Given my current unfamiliarity with Rust and it's build and library ecosystem and the size of this PR, my current plan is to mostly follow a black box testing approach:
- Review the CI changes to verify tests are running correctly
- Review changes to the tests to verify it tests our existing functionality
- Review dependencies to see if they are necessary and well-maintained
- Run the executable and verify that all our existing functionality works as expected
- Verify that all the CPU architectures we now build can compile and run successfully
- Try to do a quick tour of the codebase, starting with the entrypoint. May look around to see if I can find some SAST tool to help build a call graph (if I can find a good one)

gaborbernat:
> Given my current unfamiliarity with Rust and it's build and library ecosystem and the size of this PR, my current plan is to mostly follow a black box testing approach:

Note I already did most of that, and the CI validates most of it, however I'd be interested what other gaps I might have missed you'll find. Also the plan is to make it a major release as `rich` render becomes the default.

kemzeb:
Sounds good, I'll try to complete reviewing by the end of the weekend

rust/tests/public_api/common.rs
```diff
@@ -0,0 +1,210 @@
...
+from pathlib import Path
+from unittest.mock import create_autospec
+
+from packaging.version import Version
+from nab_python.lockfile import LockInput
+from nab_python.resolve import ResolutionResult
+import nab_python.resolve as resolve_module
+
+if not hasattr(resolve_module, "_pipdeptree_resolve"):
+    resolve_module._pipdeptree_resolve = resolve_module.resolve_pyproject
```
kemzeb: Wasn't able to test was the subcommand `from-index` because of this error:
```
vscode ➜ /workspaces/pipdeptree (poc/rust-meson) $ .tox/dev/bin/pipdeptree from-index --pyproject pyproject.toml 
AttributeError: module 'nab_python.resolve' has no attribute 'resolve_pyproject'
vscode ➜ /workspaces/pipdeptree (poc/rust-meson) $ .tox/dev/bin/pipdeptree -p nab_python
nab-python==0.0.9
┣━━ ✓ build required: >=1.2 installed: 1.5.0
┃   ┣━━ ✓ packaging required: >=24.0 installed: 26.2
┃   ┗━━ ✓ pyproject_hooks required: Any installed: 1.2.0
┣━━ ✓ installer required: >=0.7 installed: 1.0.1
┣━━ ✓ nab-index required: ==0.0.9 installed: 0.0.9
┃   ┣━━ ✓ packaging required: >=24.0 installed: 26.2
┃   ┣━━ ✓ truststore required: >=0.10 installed: 0.10.4
┃   ┣━━ ✓ typing_extensions required: >=4.6 installed: 4.16.0
┃   ┗━━ ✓ urllib3 required: >=2.0 installed: 2.7.0
┣━━ ✓ nab-resolver required: ==0.0.9 installed: 0.0.9
┃   ┗━━ ✓ typing_extensions required: >=4.6 installed: 4.16.0
┣━━ ✓ pyproject_hooks required: >=1.2 installed: 1.2.0
┣━━ ✓ tomli required: >=2.0 installed: 2.4.1
┣━━ ✓ tomli_w required: >=1.2 installed: 1.2.0
┗━━ ✓ typing_extensions required: >=4.6 installed: 4.16.0
vscode ➜ /workspaces/pipdeptree (poc/rust-meson) $ 
```
My only guess would be the problem line is here, though I'm unsure.

kemzeb (commented):
From multiple runs of the flags, I haven't observed any glaring issues. My only issue was with `from-index` which is mentioned in my review. Some issues uncovered:

1. We are missing the resolved python message:
```
vscode ➜ /workspaces/pipdeptree (poc/rust-meson) $ .tox/dev/bin/pipdeptree -p 
# <--- (resolved python: ...) should be here
pipdeptree -d 1
pipdeptree==4.0.0
┣━━ ✓ nab-index required: >=0.0.8 installed: 0.0.9
┗━━ ✓ nab-python required: >=0.0.8 installed: 0.0.9
vscode ➜ /workspaces/pipdeptree (poc/rust-meson) $ 
```

2. When using `--summary`, if the aggregated license row becomes too big, it appears that the overall summary table becomes unreadable. I think the following is happening:
- Old code used `License-Expression` and `Classifier` to get the license
- New code uses `License`, which may accept the entire license agreement as text

Currently been having trouble figuring out how to make builds for MacOS and Windows, will need to spend time learning about the rust build system and cross-compilation but don't want that to hold up the PR.

rust/tests/public_api/common.rs
gaborbernat: Two good bugs caught; pushed fix, as for the other: the message only prints when pipdeptree auto-detects an interpreter (VIRTUAL_ENV/CONDA_PREFIX/poetry). You ran ran .tox/dev/bin/pipdeptree directly, which doesn't set VIRTUAL_ENV, so nothing was detected.

PR #646, opened 2026-07-31
👷 ci(dist): pre-install Rust toolchain for aarch64 wheels

gaborbernat:
The aarch64 wheel jobs fail intermittently while building the Rust extension, with `error: failed to install component: 'clippy-preview-aarch64-unknown-linux-{gnu,musl}', detected conflict: 'bin/cargo-clippy'`. `rust-toolchain.toml` declares the `clippy`, `rustfmt`, and `llvm-tools-preview` components, so the first `cargo` call in each manylinux and musllinux container provisions the toolchain on demand, and that incremental component install races and conflicts on the `cargo-clippy` hardlink inside the aarch64 containers. Those components exist for the Rust CI job, not the wheel build, which only needs `cargo` and `rustc`.

The cibuildwheel Linux setup now installs the pinned `1.97` toolchain together with its components in the single rustup-init call, so every build finds a complete toolchain and never triggers the on-demand reinstall that conflicts. The channel and component list mirror `rust-toolchain.toml`; a comment flags that they must stay in sync.

This only touches `CIBW_BEFORE_ALL_LINUX`, so the macOS and Windows toolchain setup is unchanged.

### DIFF
--- a/.github/workflows/dist.yaml
+++ b/.github/workflows/dist.yaml
@@ -33,8 +33,11 @@ jobs:
         uses: pypa/cibuildwheel@294735312765b09d24a2fbec22660ce817587d55 # v4.1.0
         env:
           CIBW_ARCHS: ${{ matrix.arch }}
+          # Install the pinned toolchain and its components up front. Left to rust-toolchain.toml, the first
+          # cargo call in each container provisions them on demand, which races and conflicts on the
+          # cargo-clippy hardlink on aarch64. Keep the channel and components in sync with rust-toolchain.toml.
           CIBW_BEFORE_ALL_LINUX: >-
-            curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --profile minimal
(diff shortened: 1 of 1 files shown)


kemzeb:
Looks like 4.1.1 and 4.1.2 releases failed to deploy to pypi because [VERSION](https://github.com/tox-dev/pipdeptree/blob/e0b8f185d8c7265789047d7a4d614efafae3999f/VERSION) is still at 4.1.0. Going forward I see two solutions to this:
* Bump the version in the `VERSION` file and make a commit before cutting a new release
* Have the `VERSION` file be an artifact created during `dist` workflow and write the new tag name to this file

For the releases that were cut... not sure what the acceptable choice is as they were never actually pushed to pypi. Would it be acceptable to release 4.1.3 using one of the solutions above and ignore the failed releases? Or, do we delete the failed releases and release it as 4.1.1?

gaborbernat:
@kemzeb https://github.com/tox-dev/pipdeptree/releases/tag/4.2.0 should fix it

PR #660, opened 2026-08-11
ci: run Cargo and Rust toolchain Dependabot updates weekly

kemzeb:
Adds cargo and rust-toolchain ecosystems for dependabot to create updates for.


kemzeb:
dependabot provides other configurations  (e.g. pull request limits, labels) but we can make changes if the need arises

Issue #681, opened 2026-08-20
nab libraries reorganized with 0.0.14: no more nab-python


ViktorHaag:
### Describe the feature

With [v0.0.14 of nab](https://github.com/notatallshaw/nab/releases/tag/v0.0.14), they've reorganized to remove `nab-python`. As a result, greedy upgrading of one's python package environment can get fragile as there's a risk that bits of nab can get upgraded to 0.0.14 and bits stuck on 0.0.13.

It'd be useful if pipdeptree removed its dependency on nab-python, or pinned to nab==0.0.13 while preparing to adapt to the changes introduced with 0.0.14.

PR #683, opened 2026-08-20
🐛 fix(index): adopt nab-project

gaborbernat:
Viktor Haag reported in [#681](https://github.com/tox-dev/pipdeptree/issues/681) that Nab's v0.0.14 release removed `nab-python`; greedy upgrades can leave mixed Nab versions in an environment. 🐛

I replaced the dependency and direct imports with `nab-project`, which contains the project configuration and resolution APIs. The Rust workflow installs `nab-project`, keeping its public-API test environment aligned with the application dependency.

`from-index` keeps its interface and names `nab-project` in the missing-dependency error.


digitalnumeric:
What version of pipdeptree will this change appear in?  Is there a timeline for when it will appear?

gaborbernat:
Released.

PR #700, opened 2026-09-05
fix: Fix from-index resolution with nab 0.0.16

### LINKED ISSUE #697: The from-index subcommand requires nab-index and nab-project (but they are installed already!)

4.2.3

No

### Describe the problem

Running `pipdeptree i pipdeptree` yields the following error message:
```
The from-index subcommand requires nab-index and nab-project
```

but the packages are already installed; result from `pip list`, summarized:
```
Package                   Version
------------------------- -----------------
...
build                     1.6.0
...
installer                 1.0.1
...
nab-index                 0.0.16
nab-markersets            0.0.16
nab-project               0.0.16
nab-provider              0.0.16
nab-resolver              0.0.16
...
packaging                 26.3
...
pipdeptree                4.2.3
...
pyproject_hooks           1.2.0
...
tomli                     2.4.1
tomli_w                   1.2.0
...
truststore                0.10.4
...
typing_extensions         4.16.0
...
urllib3                   2.7.0
...
```

It seems that the error comes from `import nab_project.config`, it no longer works as of `nab-project` version 0.0.16.  
I managed to reproduce this problem on macOS Tahoe 26.6.2 and Fedora 44.

kemzeb:
Attempts to fix #697.

Current approach is to bring over the `nab` package that holds the original API we were using that was moved to this package.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -30,7 +30,7 @@ jobs:
         with:
           enable-cache: false
       - name: Install resolver dependencies
-        run: uv pip install --target .python-packages 'nab-index>=0.0.11' 'nab-project>=0.0.14'
+        run: uv pip install --target .python-packages 'nab==0.0.16'
       - name: Check formatting
         run: cargo fmt --all --check
       - name: Check lints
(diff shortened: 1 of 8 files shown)


kemzeb:
Making it a draft PR, since this is a first for me in using a coding agent and want more time to review it (exciting stuff so far). But I also want to show one approach to this problem.

I'm not sure if we should use `nab` since it's the CLI utility. Then again it does expose `nab.config.model` as public API. Since were looking to fully support `[tool.nab]` in `pyproject.toml` this would be the easiest solution. Alternatively, we could see what else we could do with the existing libraries. This may involve us having to maintain our own logic for parsing `[tool.nab]`.

gaborbernat:
Superseeded by https://github.com/tox-dev/pipdeptree/pull/704

